Issue #4, opened 2018-09-17
Feature request: impl Clone


starkat99:
impl Clone for OnceCell<T> where T: Clone

Finding myself in need of Cloning. I'm guessing Lazy can't be, but OnceCell would be helpful?

matklad:
unsync::OnceCell is now Clone

mmstick:
Would it be reasonable to support this for `sync::OnceCell` as well?

mmstick:
I guess `Arc<OnceCell>` achieves this.

PR #5, opened 2018-09-19
impl Clone for OnceCell<T>

matklad:
Closes https://github.com/matklad/once_cell/issues/4

@starkat99 one thing I am worried about that there's no *strong* semantics here. What happens if one thread is calling `get_or_init`, and the over thread is calling `.clone` is indeterminate: you might get either an empty or a full cell. I think that's the reason why synch primitives like `Once` or `Mutex` don't implement `Clone` in general? Will this semantics work for you?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- impl `Clone` for `OnceCell<T>` if `T: Clone>
 
 ## 1.0.5
- - No changelog until this point :(
+- No changelog until this point :(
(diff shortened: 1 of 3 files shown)


starkat99:
Oh sorry, I should have been more specific, I didn't realize the sync version was the same name. I agree, semantics are very unclear for sync version, so maybe don't need to do that? I was just needing single thread version to be clone.

matklad:
unsync case is totally uncontroversial, implemented in https://github.com/matklad/once_cell/commit/084ffbdd5ee76a1fac1a78e6bdff0b23e18611a6 and relased as 1.0.6

PR #7, opened 2019-01-11
Add PartialEq, From<T> derives + Fix rustc 1.24.1 builds

mmstick:
I'd like to use this within System76 / Pop!_OS projects, but that requires an extra derive (PartialEq), and using `use` import syntax that's compatible with rustc 1.24.1. While I was there, added some additional missing derives (Hash, From<T>).

### DIFF
--- a/src/imp_pl.rs
+++ b/src/imp_pl.rs
@@ -1,6 +1,7 @@
 extern crate parking_lot;
 
 use std::cell::UnsafeCell;
+use std::hash::{Hash, Hasher};
 use self::parking_lot::{Once, ONCE_INIT, OnceState};
 
 /// A thread-safe cell which can be written to only once.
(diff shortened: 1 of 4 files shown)


matklad:
Hm, I think we defenetely can’t implement Hash: otherwise, you can store an empty cell in a hash map, then fill it, and get a hash map in an inconsistent state.

It can be said that Eq and Ord could screw data structures in a similar way, but I think I’d be fine with those. Cc 
https://github.com/rust-lang/rust/pull/33306

mmstick:
I've removed the Hash derives.

src/imp_pl.rs
```diff
@@ -38,6 +38,20 @@ impl<T> Default for OnceCell<T> {
...
+        cell
+    }
+}
+
+impl<T: PartialEq> PartialEq for OnceCell<T> {
+    fn eq(&self, other: &OnceCell<T>) -> bool {
+        self.get() == other.get()
+    }
+}
+
```
matklad: This probably needs to be duplicated in impl_std?

And perhaps it makes sense to add a trivial test for this API as well, exactly to test that both impls in fact have the same api?

matklad:
Awesome, thanks!

matklad:
released as 0.1.7

PR #9, opened 2019-02-07
Update crossbeam-utils to ^0.6.0

khardix:
I'm currently attempting to package this crate for [Fedora](https://fedoraproject.org) which already contains version `0.6.3` of the `crossbeam-utils` crate; in order to do so, the dependencies have to be updated to use compatible version.

The changes needed are fairly small (AFAIK only function signature changes and only in test suite), so I went ahead and prepared a patch.

Let me know if there is anything else I can do to help.

### DIFF
--- a/tests/test.rs
+++ b/tests/test.rs
@@ -214,7 +214,7 @@ fn sync_once_cell_does_not_leak_partially_constructed_boxes() {
         let cell: sync::OnceCell<String> = sync::OnceCell::INIT;
         scope(|scope| {
             for _ in 0..n_readers {
-                scope.spawn(|| loop {
+                scope.spawn(|_| loop {
                     if let Some(msg) = cell.get() {
                         assert_eq!(msg, MSG);
                         break;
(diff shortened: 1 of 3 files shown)


tests/test.rs
```diff
@@ -214,17 +214,17 @@ fn sync_once_cell_does_not_leak_partially_constructed_boxes() {
...
                         break;
                     }
                 });
             }
             for _ in 0..n_writers {
-                scope.spawn(|| cell.set(MSG.to_owned()));
+                scope.spawn(|_| cell.set(MSG.to_owned()));
             }
-        })
+        }).unwrap()
```
khardix: Note that `scope()` now returns `Result<R>` which is an `Err` if any of the scoped threads panicked.
`unwrap`ing that should preserve the previous behaviour (propagating thread panic).

matklad:
Thanks!

bors r+

matklad:
bors ping

matklad:
published as 1.0.8

PR #15, opened 2019-05-10
remove macro_use in code and comments

ldesgoui:
They're not required anymore so I removed them

Thanks for the neat crate


matklad:
Indeed, thanks!

bors r+

PR #25, opened 2019-06-07
Specify MSRV policy

matklad:
cc @BurntSushi

I think I can do a neat trick of guaranteeing conservative MSRV if features are disabled. 

With `parking_lot` feature enabled, I am bound to at least that crate's MSRV, so I can't make promises.

Additionally, at some point, when `std::sync::Mutex::new` becomes `const`, I'd like to take advantage of that. I plan to add a *default* `std_once_cell_get_or_try_init` feature at that moment, which will bump MSRV *if* you use default features. 

Also, the crate's version is currently 0.2, but, API-wise, I am ready to do 1.0. However I'd prefer to do that later (months?) lest someone finds a critical issue with the current interface.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 might store arbitrary non-`Copy` types, can be assigned to at most once and provide direct access
 to the stored contents. In a nutshell, API looks *roughly* like this:
 
-```rust,no-run
+```rust,ignore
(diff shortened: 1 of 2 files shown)


BurntSushi:
Awesome. Sounds great. Thank you. :)

Issue #28, opened 2019-07-05
Feature request: Add OnceCell::try_unwrap()


manuthambi:
Add a method which consumes OnceCell and returns the element if it exists.
The method signature can be very similar to Arc::try_unwrap().

`pub fn try_unwrap(self) -> Result<T, OnceCell<T>>`

manuthambi:
On second thought, there isn't much point in returning an uninitialized OnceCell if the unwrap fails.
So the signature could be:
`pub fn try_unwrap(self) -> Option<T>`

matklad:
sgmt, though, I think the name and signature should be, to map `RefCell`:

```
pub fn into_inner(self) -> Option<T>
```

manuthambi:
`into_inner` sounds better, and it is better to be consistent with `RefCell`.

I haven't contributed to Rust before, but do you want me to send you a pull request?

matklad:
Sure, that would be awesome! Here's a checklist of things to do:

- [ ] add the method to the `unsync` variant
- [ ] add the method to both versions of the `sync` variant
- [ ] write a short docstring for them, with a doc test
- [ ] write a simple smoke test

PR #29, opened 2019-07-05
Added OnceCell::into_inner() for sync and unsync.

manuthambi:
#28

### DIFF
--- a/src/imp_pl.rs
+++ b/src/imp_pl.rs
@@ -87,7 +87,7 @@ impl<T> OnceCell<T> {
         }
     }
 
-    pub fn get_or_try_init<F: FnOnce() -> Result<T, E>, E>(&self, f: F) -> Result<&T, E> {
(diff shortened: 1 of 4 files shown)


src/imp_pl.rs
```diff
@@ -87,7 +87,7 @@ impl<T> OnceCell<T> {
         }
     }
 
-    pub fn get_or_try_init<F: FnOnce() -> Result<T, E>, E>(&self, f: F) -> Result<&T, E> {
+    pub(crate) fn get_or_try_init<F: FnOnce() -> Result<T, E>, E>(&self, f: F) -> Result<&T, E> {
```
matklad: both `get_or_try_init` and `into_inner` should be just `pub` to be visible from outside

matklad: Ah, sorry, that's pl impl, it should be `pub(crate)` indeed, thanks!

manuthambi:
Thanks for reviewing and merging it!

matklad:
published as 0.2.2!

Issue #1, opened 2018-08-02
Unsafe audit


matklad:
This crate uses `unsafe` quite a bit for both `sync` and `unsync` variants of `OnceCell`. I feel rather confident about `unsync` usages of unsafe: the implementation is the same as in the `lazycell`. However the `sync` version, which is inspired by `lazy_static`, has quite a bit more of my own invention.

It would be cool if knowledgeable people in the area reviewed the code for safety issues!

matklad:
@kohensu I initially thought just like that, but now I am thinking that SlipperyFrob is, in fact, 100% correct. Here's a manually forced reordering which I think would be correct under `Relaxed`, and which actually makes the test fail:

https://github.com/matklad/once_cell/blob/0c7c8e7194e312d0d45664f340bb9da57b0e3094/src/lib.rs#L504-L513

The problem is, the thread that calls get does not call `set_inner`, so the `call_once` synchronization has no effect on it.

kohensu:
@matklad, my bad, I was focusing on the `get_or_init()` part.
Indeed there was a race between the `get()` and the `set()`.
By the way thank you for taking the time to correct me.

matklad:
I've significantly reworked sync version internals. Now, for parking lot case it leverages `Once::state` method for synchronization. For std version, we have to maintain `is_initialized` flag ourselves unfortunately.

matklad:
I think this isn't actionable at this point: we are already deployed throughout the ecosystem :)

And, well, we now have some extensive miri tests, which inspire at least some confidence!

PR #33, opened 2019-07-15
Fix unsync unsound aliasing of Option<T>

danielhenrymantilla:
As per https://github.com/matklad/once_cell/issues/1, I have looked at `unsync::OnceCell` code, and found that `::once_cell::unsync::OnceCell::set` is unsound when combined with `::once_cell::unsync::OnceCell::get`:

  - for a brief moment, before `Err`-oring, we have a unique reference to the inner `Option<T>`, despite it being potentially aliased by a previous `.get()` borrow.

For instance, miri rejects the following code:

```rust
fn main ()
{
    let x = ::once_cell::unsync::OnceCell::new();
    x.set(42).unwrap();
    let at_x = x.get().unwrap(); // --- (shared) borrow of inner `Option<T>` --+
    let _ = x.set(27); // <-- temporary (unique) borrow of inner `Option<T>`   |
    println!("{}", at_x); // <------- up until here ---------------------------+
}
```

Changing the kind of borrow with which the discriminant of `Option<T>` is read to be _shared_, before getting a unique borrow only on the non-`Err`-oring path fixes the issue, as done in this PR.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -279,10 +279,11 @@ pub mod unsync {
         /// assert!(cell.get().is_some());
         /// ```
         pub fn set(&self, value: T) -> Result<(), T> {
-            let slot = unsafe { &mut *self.inner.get() };
+            let slot = unsafe { &*self.inner.get() };
             if slot.is_some() {
                 return Err(value);
             }
+            let slot = unsafe { &mut*self.inner.get() };
             // This is the only place where we set the slot, no races
             // due to reentrancy/concurrency are possible, and we've
             // checked that slot is currently `None`, so this write


matklad:
This is tricky, but makes sense. I wonder if there's some kind of more direct way to get this "upgradable read" behavior: in the current code, doing the second "load" of `slot` is basically a no-op anyway, but I can imagine situations where fetching a refernce again causes overhead...

Did you manage to run the test suite under miri somehow? If I try to do this locally, I get ` Miri evaluation error: miri does not support dynamically loading libraries`

bors r+

Thanks!

danielhenrymantilla:
> Did you manage to run the test suite under miri somehow? If I try to do this locally, I get Miri evaluation error: miri does not support dynamically loading libraries

Same for me: the `sync` / multithreading / `pthread`-based stuff seems to use something that is not supported by miri.

Shnatsel:
Can this lead to memory corruption or concurrency issues in practice?

If so, please file a security advisory at https://github.com/RustSec/advisory-db so that dependents of this crate can check if they're running a vulnerable version and upgrade.

matklad:
I don’t think so: this violates stacked borrows model, but compiler most likely does’t exploit this yet. 

In other words, this is unsound according to a memory model which is not formalized yet, and no known miscompilations exist.

matklad:
@Shnatsel not that advisory will be filed due to https://github.com/matklad/once_cell/issues/46

PR #34, opened 2019-07-15
Update parking_lot

kjeremy:
While this isn't strictly necessary I'm trying to remove multiple versions of rand from downstream.


matklad:
Excellent! Could you replace explicit parking_lot_api dep with a reexport as well? See https://github.com/matklad/once_cell/pull/27/files

matklad:
bors r+

Thanks!

PR #37, opened 2019-08-07
changed `Lazy` to use `FnOnce` instead of `Fn`

KrishnaSannasi:
This change allows for more ergonomic usage of `Lazy` when embedded in other types or on the stack (not a global `static`). It does this by allowing strictly more types of closures to be used with `Lazy`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -190,6 +190,7 @@ pub mod unsync {
         ops::Deref,
         cell::UnsafeCell,
         panic::{UnwindSafe, RefUnwindSafe},
+        hint::unreachable_unchecked
     };
 
     /// A cell which can be written to only once. Not thread safe.
@@ -453,7 +454,7 @@ pub mod unsync {
             this.cell.get_or_init(|| unsafe {
                 match (*this.init.get()).take() {
                     Some(f) => f(),
-                    None => std::hint::unreachable_unchecked()
+                    None => unreachable_unchecked()
                 }
             })
         }
@@ -469,8 +470,11 @@ pub mod unsync {
 
 pub mod sync {
     use crate::imp::OnceCell as Imp;
-    use std::fmt;
-    use std::cell::UnsafeCell;
+    use std::{
+        fmt,
+        cell::UnsafeCell,
+        hint::unreachable_unchecked
+    };
 
     /// A thread-safe cell which can be written to only once.
     ///
@@ -741,7 +745,7 @@ pub mod sync {
             this.cell.get_or_init(|| unsafe {
                 match (*this.init.get()).take() {
                     Some(f) => f(),
-                    None => std::hint::unreachable_unchecked()
+                    None => unreachable_unchecked()
                 }
             })
         }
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -448,7 +448,14 @@ pub mod unsync {
         /// assert_eq!(&*lazy, &92);
         /// ```
         pub fn force(this: &Lazy<T, F>) -> &T {
-            this.cell.get_or_init(|| (this.init)())
+            // Safe because closure is guaranteed to be called at most once
+            // so we only call `F` once, this also guarantees no race conditions
```
matklad: Interesting! Let's add this to docs of `get_or_init`? Something like this:

> Several threads can call `get_or_init` concurrently, but at most one closure will be executed.

There *can* be an impl that eagerly calls the closure and that uses `.set`. It's useful to say that the current impl guranted not to do that

src/lib.rs
```diff
@@ -448,7 +448,14 @@ pub mod unsync {
...
+            this.cell.get_or_init(|| unsafe {
+                match (*this.init.get()).take() {
+                    Some(f) => f(),
+                    None => std::hint::unreachable_unchecked()
```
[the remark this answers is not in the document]
KrishnaSannasi: Done

src/lib.rs
KrishnaSannasi: That already seems to be part of the docs!
> ```rust
> ...
> /// Gets the contents of the cell, initializing it with `f`
> /// if the cell was empty. May threads may call `get_or_init`
> /// concurrently with different initializing functions, but
> /// it is guaranteed that only one function will be executed.
> ...
> ```

matklad: :rofl: I was reading the unsync docs

matklad:
bors r+

Thanks!

matklad:
published as 0.2.5

KrishnaSannasi:
Ahhhh! Just realized I forgot to update the `Deref` impls.

matklad:
Would you like to send a follow up PR? If you bump version in Cargo.toml and add an entry to changelog, that would be double awesome :)

KrishnaSannasi:
Done

src/lib.rs
```diff
@@ -683,21 +695,37 @@ pub mod sync {
...
+        }
+    }
+
+    // We never create a `&F` from a `&Lazy<T, F>` so it is fine
+    // to not impl `Sync` for `F`
+    // we do create a `&mut Option<F>` in `force`, but this is
+    // properly synchronized, so it only happens once
+    // so it also does not contribute to this impl
+    // We do create a `&T` from `&Lazy<T, F>`, so `T` needs `Sync`
+    unsafe impl<T: Sync, F: Send> Sync for Lazy<T, F> {}
```
matklad: I've missed two things here I think:

* a `Send` impl is missing
* we actually need `T: Send + Sync`  for `Sync`: If we create `Lazy` in thread1, call `force` in thread2, and drop the `Lazy` in thread1, we effectively move `T` from thread1 to thread2. 

I'll fix both

Issue #53, opened 2019-09-01
Provide non blocking subset of sync::OnceCell for no_std


matklad:
OnceCell::get and OnceCell:: set could work without blocking, so we should support these APIs in no std. Note that we intentionally do not want to implement no_std blocking via spinning, see https://www.reddit.com/r/rust/comments/cy910u/comment/eyrfmna for rationale.

matklad:
Thinking of it, looks like even `set` can't be lock, in general? You need to atomically set *both* the is_completed state and the value, and you can only do that if you can pack both into a single word.

You can do that if you heap-allocate the value, but that again needs std.

pitdicker:
> Thinking of it, looks like even set can't be lock, in general? You need to atomically set both the is_completed state and the value, and you can only do that if you can pack both into a single word.

Would one atomic with three states not be enough?
```rust
pub struct OnceCell<T> {
    status: AtomicUsize,
    value: UnsafeCell<Option<T>>,
}

pub fn set(&self, value: T) -> Result<(), T> {
    match self.state.load(Ordering::Aquire) {
        UNINITIALIZED => {
            let old = self.state.compare_and_swap(UNINITIALIZED, RUNNING, Ordering::Relaxed);
            if old != UNINITIALIZED {
                // Another thread must have initialized the cell just now, or is in the process of initializing it.
                Err(value)
            }
            let slot = unsafe { &mut *self.inner.get() };
            *slot = Some(value);
            self.state.store(INITIALIZED, Ordering::Release);
            Ok(())
        },
        RUNNING | INITIALIZED => Err(value),
    }
}
```

I suppose that if the initializing thread gets suspended while the state is `RUNNING`, all threads are out of luck until it can continue, but don't really see other problems yet.

matklad:
Hm, that’s interesting!

I think we perhaps need two methods with set-like semantics: one that might block, but guarantees that the cell is initialized (maybe even returns an &T), and one that does not block, bit also does not guarantee that the cell is initialized.

pitdicker:
Changing the current `set` to not block would be a difficult change, as code currently could call `get_unchecked` after a `set` and expect it to be safe.

A non blocking `set` would mean the first `get` has to be used in a loop or something until it succeeds. I see not blocking as less then optimal, but suppose such a new method can unlock some uses and support no_std.

👍 to have `set` return &T

matklad:
This is now more or less availble in the `race` module

PR #54, opened 2019-09-02
Allow non-std version of crate to be used in no_std environments

YuhanLiin:
This changes makes the crate no_std by default and brings in the std crate as an optional dependency. This allows the `unsync` part of the crate to be useable in no_std environments when the `std` feature is disabled.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -221,6 +221,11 @@ might be easier to debug than a deadlock.
 
 */
 
+#![no_std]
+
+#[cfg(feature = "std")]
+extern crate std;
+
 #[cfg(feature = "std")]
 #[cfg(feature = "parking_lot")]
 #[path = "imp_pl.rs"]


matklad:
Heh, it's amazing how I managed to correctly puge every std dependency without actually enabling no_std. I think I'll do a different impl:

```
#![cfg_attr(not(feature = "std"), no_std)]
```

Am I correct that this is equivalent to your PR?

matklad:
I'll do a minor release with this PR merged when the first of this happens:

* someone explicitly asks for a release with this feature
* https://github.com/matklad/once_cell/issues/53 is fixed
* there's an unrelated reason to do a minor release

matklad:
#53 is unfortunatelly not fixable, so, released 1.0.2

YuhanLiin:
Yeah the code you posted does the same thing.

PR #56, opened 2019-09-02
impl Default for Lazy

alexheretic:
Replacing lazy_static in my codebases I've found I'm doing `Lazy::new(<_>::default)` a fair bit generally for Mutex use.

This pr streamlines that to `Lazy::default()`, in a similar way to `Arc`, `Mutex` wrappers. Although it won't work in static contexts, which sucks.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -537,6 +537,13 @@ pub mod unsync {
             Lazy::force(self)
         }
     }
+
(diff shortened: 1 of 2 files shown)


matklad:
Makes sense to add, yeah!

bors r+

matklad:
published as 1.1.0

Issue #59, opened 2019-09-11
Please allow to skip checks with unsafe methods


glandium:
It would be useful to have unsafe methods that allow to get to the underlying values in OnceCell and Lazy, that assume initialization happened. A use case would be to avoid the initialization check when you know by other means that initialization did happen.

matklad:
Yeah, something like this would make sense:

```
/// Safety: the cell must be in initialized state. 
unsafe fn get_unchecked(&self) -> &T
```

Could you give a specific example where this is useful though? My gut feeling is that, to use this function safely, you need to prove that initialization indeed has happened. A natural proof would be the `&T` you get from calling the initialization function. However, if you can thread a `&T` to the `get_unchecked` callsite, you don't need `get_unchecked`! And if you can't thread the `&T`, there's a high chance that the cell might not be initialized.

matklad:
Thanks for the example! I agree that this is something which can't be expressed using only safe code, without compromising elsewhere. 

Specifically, `FooHandle` *can* be defined as `FooHandle(&'static Foo)`, and that would make this pattern fully-safe, at the expense of making `FooHandle` a non-ZST type.  If we want a ZST, we need get unchecked. BTW, another way to define `FooHandle` is `FooHandle(PhantomData<&'static Foo>)`, which maybe is better with respect to auto-traits.

My plan is to add `OnceCell::get_unchecked` but not `Lazy::get_unchecked`: `Lazy` is a purely convenience type, and I fear that adding unsafe methods to it might lead to misuse. So, the example would look like this:

```rust
pub struct Foo { ... }

#[derive(Copy, Clone)]
pub struct FooHadle(PhantomData<&'static Foo>);

static INSTANCE: sync::OnceCell<Foo> = sync::OnceCell::new();

impl Foo {
    fn new() -> Self { ... }
    fn new_handle() -> FooHandle {
        INSTANCE.get_or_init(Foo::new);
        FooHandle(PhantomData)
    }
}

impl FooHandle {
    fn get(self) -> &'static Foo {
        unsafe { INSTANCE.get_unchecked() }
    }
}

```

glandium:
Heh, I oversimplified my actual use-case, which does, in fact, have `PhantomData<&'static Foo>`.

xfix:
Hm, cannot `.get().unwrap_or_else(|| std::hint::unreachable_unchecked())` be used for that purpose? I think LLVM should be able to optimize that well.

matklad:
[...]

PR #60, opened 2019-09-14
add sync::OnceCell::get_unchecked

matklad:
Note that we intentionally don't add
`unsync::OnceCell::get_unchecked`:

* it doesn't do any atomic operations, so should be cheap
* given than it's cheap, there's a chance that `get_unchecked` will be
  misused
* if you absolutely need it, you can `.unwrap_unchecked` on the call
  site

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.2.0
+
+- add `sync::OnceCell::get_unchecked`.
+
 ## 1.1.0
 
 - implement `Default` for `Lazy`: it creates an empty `Lazy<T>` which is initialized with `T::default` on first access.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "once_cell"
-version = "1.1.0"
+version = "1.2.0"
 authors = ["Aleksey Kladov <aleksey.kladov@gmail.com>"]
 license = "MIT OR Apache-2.0"
 edition = "2018"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -656,11 +656,14 @@ pub mod sync {
             unsafe { &mut *self.0.value.get() }.as_mut()
         }
 
-        /// Safety to call if guarded by `initialize`, `is_initialized`
+        /// Get the reference to the underlying value, without checking if the
+        /// cell is initialized.
         ///
-        /// Implementations of those functions in `imp` must provide proper
-        /// synchronization and write-once property
-        unsafe fn get_unchecked(&self) -> &T {
+        /// Safety:
+        ///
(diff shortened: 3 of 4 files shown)


matklad:
I believe this is correct semantically. However, I am worry about social consequences here: by having this method, we create a risk of this method being misused. It's unclear to me if potential performance benefits (which are real, see #59) outweigh the risk of misuse.

matklad:
I did a quick experiment to check if `std::hint::unrecheable_unchecked` can be used on the call site to eliminate the load. It looks like it doesn't help :(


```rust
#[inline(never)]
#[no_mangle]
pub fn quux_checked(cell: &OnceCell<String>) -> &String {
    cell.get().unwrap()
}

#[inline(never)]
#[no_mangle]
pub fn quux_unchecked(cell: &OnceCell<String>) -> &String {
    unsafe { cell.get_unchecked() }
}

#[inline(never)]
#[no_mangle]
pub fn quux_manual_unchecked(cell: &OnceCell<String>) -> &String {
    unsafe {
        match cell.get() {
            Some(value) => value,
            None => std::hint::unreachable_unchecked(),
        }
    }
}
```


```asm
0000000000000000 <quux_checked>:
   0:   50                      push   rax
   1:   48 8b 0f                mov    rcx,QWORD PTR [rdi]
   4:   48 83 f9 02             cmp    rcx,0x2
   8:   75 09                   jne    13 <quux_checked+0x13>
   a:   48 89 f8                mov    rax,rdi
   d:   48 83 c0 08             add    rax,0x8
  11:   59                      pop    rcx
  12:   c3                      ret    
  13:   48 8d 3d 00 00 00 00    lea    rdi,[rip+0x0]        # 1a <quux_checked+0x1a>
  1a:   ff 15 00 00 00 00       call   QWORD PTR [rip+0x0]        # 20 <quux_checked+0x20>
  20:   0f 0b                   ud2    

Disassembly of section .text.quux_unchecked:

0000000000000000 <quux_unchecked>:
   0:   48 8d 47 08             lea    rax,[rdi+0x8]
   4:   c3                      ret    

Disassembly of section .text.quux_manual_unchecked:

0000000000000000 <quux_manual_unchecked>:
   0:   48 8b 0f                mov    rcx,QWORD PTR [rdi]
   3:   48 83 c7 08             add    rdi,0x8
   7:   31 c0                   xor    eax,eax
   9:   48 83 f9 02             cmp    rcx,0x2
   d:   48 0f 44 c7             cmove  rax,rdi
  11:   c3                      ret    
```

matklad:
bors r+

CI failied due to miri failure on nightly, let's see if its any better now...

matklad:
merging manually, as miri failure seems unrelated. 

I am still not sure if it is definitelly worth it, but, becuase the method is usnafe and makes sense, I feel like it's ok to merge this.

PR #62, opened 2019-10-18
Allow reentrant initialization of unsync::OnceCell

pitdicker:
This is not really a serious pull request, but I am interested in your thoughts.

I don't think it is necessary for the unsync variant of `OnceCell` to panic on reentrant initialization: because there are no other references to its interior when the result is written by `get_or_init`, the double write can not be observed outside the initializer, and no aliasing guarantees are violated.

Still a panic may be useful because this may indicate an error, and because it gives some symmetry to the sync variant.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -373,13 +373,13 @@ pub mod unsync {
         /// Gets the contents of the cell, initializing it with `f`
         /// if the cell was empty.
         ///
+        /// While probably an error, it is possible to reentrantly initialize
+        /// the cell from `f`.
+        ///
         /// # Panics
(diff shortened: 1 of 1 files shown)


matklad:
I believe this is unsound, see a new test in https://github.com/matklad/once_cell/pull/63

pitdicker:
You are right, it is possible for the closure to sneak out a reference. Good test! Do you think it is useful to document the reason(s) reentrant initialization is not allowed?

matklad:
Good call: https://github.com/matklad/once_cell/pull/64

PR #66, opened 2019-10-21
Fix tests

pitdicker:
I wanted to test my changes in a local branch didn't change the size of `OnceCell`, and noticed some tests do not run. Apparently `#[path = "somepath.rs"]` does not work together with the test infrastructure.

Next the `test_size` test didn't compile, and I got some semi related commits related to figuring out the size it should be.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ exclude = ["*.png", "*.svg", "/Cargo.lock.min", "/.travis.yml", "/run-miri-tests
 # Uses parking_lot to implement once_cell::sync::OnceCell.
 # This makes not speed difference, but makes each OnceCell<T>
 # for up to two bytes smaller, depending on the size of the T.
-parking_lot = { version = "0.9.0", optional = true, default_features = false }
+parking_lot = { version = "0.9.0", optional = true }
 
 [dev-dependencies]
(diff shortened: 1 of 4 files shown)


matklad:
Auch, it's not `#[path]`, it's that `cfg(pointer_width = 64)` is not a thing, the right syntax is `cfg(target_pointer_width = 64)`. I've made an alternative PR, https://github.com/matklad/once_cell/pull/67, with I hope more targeted changes, but looks like it's going to fail on old rust?

matklad:
Wow, that PR passed, and that I think means that we don't test pl on CI at all :( (I think we've lost this once we stopped making parking-lot a default feature). Let's see if just https://github.com/matklad/once_cell/pull/68 works...

pitdicker:
I must have been using a rust version that was too old. Also testing under rust 1.13.1 doesn't work, but as long as it builds with that version it is ok?

matklad:
> Also testing under rust 1.13.1 doesn't work, but as long as it builds with that version it is ok?

Yeah, we explicitly only check if the code builds with 1.31.1, and only without parking lot feature. 

I think this might be closed now, thanks a lot for pointing out two(!) gaping wholes in the testing infra!

PR #68, opened 2019-10-21
actually test parking_lot on CI

matklad:
bors r+


pitdicker:
With this change it will no longer test the std implementation?

matklad:
nope, this entry tests default features (and std is default): https://github.com/matklad/once_cell/pull/68/files#diff-354f30a63fb0907d4ad57269548329e3L16-L19

PR #70, opened 2019-11-01
Add synchronization note to get_unchecked

pitdicker:
The caller not only has to make sure that the cell is initialized, but also that the data is acquired by the current thread. I can't imagine any scenario where you know it is initialized but where you didn't do an acquire yet, but it seems good to document.


matklad:
bors r+

Thanks!

PR #72, opened 2019-11-02
Use MaybeUninit

pitdicker:
Replace `UnsafeCell<Option<T>>` with `UnsafeCell<MaybeUninit<T>>`.

I included a terrible imitation of `MaybeUninit` based on `Option`, so there is no change in behavior right now. The real implementation that uses `std::mem::MaybeUninit` is behind a feature flag `maybe_uninit` and only usable with nightly.

I did find one complexity: `OnceCell` now has to implement `Drop` itself. Therefore it has to check `is_initialized`. `into_inner` now has to use the `mem::replace` trick to move the `value` field out of the `OnceCell`, and then free the cell without dropping (because `initialized` is not reset).

Conclusion: a bit messy. But I think it makes sense, as a preparation for the future. And it is a plus that `is_initialized` is now the single source of truth.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,11 @@ matrix:
     script:
     - cargo build --no-default-features --test test
 
+  - rust: stable
+    script:
+    - cargo test --no-default-features --features "std maybe_uninit"
+    - cargo test --no-default-features --features "std maybe_uninit parking_lot"
+
   - rust: 1.31.1
     script:
     - mv Cargo.lock.min Cargo.lock
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,8 @@ regex =  "1.2.0"
 default = ["std"]
 # Enables `once_cell::sync` module.
 std = []
+# Use `MaybeUninit` in the `sync` module.
+maybe_uninit = []
 
 [[example]]
 name = "bench"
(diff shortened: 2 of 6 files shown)


pitdicker:
It probably works better to swap the types to `MaybeUninit<UnsafeCell<T>>`. Then the decision whether to get a mutable reference or not can be made later. This helps making a nicer abstraction for a consume operation, and the `maybe_uninit` feature wouldn't depend on nightly. Have to think it through a bit more.

pitdicker:
Late night ideas don't always work :smile:.
To initialize a `MaybeUninit<UnsafeCell<T>>` you either have to do:
* ```
  let slot = (&*self.value.as_ptr()).get() as *mut T;
  slot.write(value);
  ```
  which is UB, because it is creating a shared reference to an uninitialized type
* ```
  let slot = self.value.as_mut_ptr();
  slot.write(UnsafeCell::new(value));
  ```
  which is not possible, because we don't have a mutable reference to the `MaybeUninit`.

The mention of `UnsafeCell` in the description [`MaybeUninit::as_ptr`](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#as_ptr.v) threw me off.

matklad:
Turns out we can't use MaybeUninit -- we'll need `#[may_dangle]` for dropcheck, and it only is avable in nightly.

PR #75, opened 2019-11-04
Use internal mutability in Waiter

pitdicker:


pitdicker:
Let's wait a little longer with this one, it turns out not to be enough.

matklad:
Is the upstream rust-lang/rust PR merged already?

pitdicker:
Yes, I think this is good to go.

src/imp_std.rs
```diff
@@ -148,32 +152,46 @@ fn initialize_inner(my_state: &AtomicUsize, init: &mut dyn FnMut() -> bool) -> b
...
-                };
-                let me = &mut node as *mut Waiter as usize;
-                assert!(me & STATE_MASK == 0);
-
-                while state & STATE_MASK == RUNNING {
-                    node.next = (state & !STATE_MASK) as *mut Waiter;
-                    let old = my_state.compare_and_swap(state, me | RUNNING, Ordering::SeqCst);
+                // Note: the following code was carefully written to avoid creating a
+                // mutable reference to `node` that gets aliased.
+                loop {
```
matklad: Can we extract this into a function, like the one in std?

matklad: purerly to keep two implementations linewise isomorphic :)

matklad:
I've just re-copy-pasted your code from std in https://github.com/matklad/once_cell/pull/81, thanks for working on this!

pitdicker:
Sorry for not following up on things!

PR #77, opened 2019-12-07
Miri now supports panics

RalfJung:
Also ignore tests that don't run in Miri instead of removing them entirely; this avoids warnings.

### DIFF
--- a/src/imp_std.rs
+++ b/src/imp_std.rs
@@ -212,7 +212,6 @@ impl Drop for Finish<'_> {
 #[cfg(test)]
 mod tests {
     use std::panic;
-    #[cfg(not(miri))] // miri doesn't support threads
     use std::{sync::mpsc::channel, thread};
 
     use super::OnceCell;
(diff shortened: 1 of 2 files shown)


tests/test.rs
```diff
@@ -413,7 +410,7 @@ mod sync {
     }
 
     #[test]
-    #[cfg(not(miri))] // leaks memory
+    #[cfg_attr(miri, ignore)] // leaks memory
```
RalfJung: Would t be useful for you to have a flag in Miri to ignore memory leaks? That would ignore leaks for all tests, but it would enable these tests to run.

matklad: I'd say in this case that's not really a leak, so ideally miri shouldn't flag it as such. But in the meantime, ignore-leaks flag would be useful, yeah!

src/imp_std.rs
```diff
@@ -235,7 +234,7 @@ mod tests {
     }
 
     #[test]
-    #[cfg(not(miri))] // miri doesn't support threads
+    #[cfg_attr(miri, ignore)] // miri doesn't support threads
```
matklad: :heart_eyes:

PR #78, opened 2019-12-09
Miri: ignore memory leaks so that we can test more things

RalfJung:


matklad:
bors r+

Thanks!

RalfJung:
I had a typo in there and you were too fast.^^ Sorry for that.

matklad:
Yeah, my approach to merging most PRs is basically r+ immediately, because

* it's unlikely that bad PR passes the tests
* in the rare event that it does breaks something, it's easier to just send a follow-up PR

^^

bors merge

PR #82, opened 2020-01-01
Add no-threads mode, for use in single-threaded environments

matklad:


src/lib.rs
```diff
@@ -228,7 +228,7 @@ might be easier to debug than a deadlock.
 #[path = "imp_pl.rs"]
 mod imp;
 
-#[cfg(feature = "std")]
+#[cfg(any(feature = "std", feature = "assume-no-threads-or-interrupts"))]
```
josephlr, line 231: It would be nice if we could just declare the `imp` module inside the `sync` module, so we don't have to worry about these `cfg`s getting out of sync.

However, https://github.com/rust-lang/rust/issues/35016 is still not fixed, so the only way to do this is moving the `imp_*.rs` files (which seems like a hassle).

matklad:
#118 is a better soln than this, available in 1.5.1

PR #88, opened 2020-01-17
Don't require Debug bound closure for Lazy

matklad:
bors r+


pitdicker:
You are quick!

matklad:
But you are the one who noticed the issue :D

PR #91, opened 2020-01-27
Upgrade parking_lot to 0.10 version and use MaybeUninit for it

tyranron:
This PR upgrades `parking_lot` dependency to `0.10` version.
- [Changelog](https://github.com/Amanieu/parking_lot/blob/0.10.0/CHANGELOG.md#parking_lot-0100-parking_lot_core-070-lock_api-032-2019-11-25)

### Additionally

Makes `OnceCell` to use `MaybeUninit` under-the-hood when `parking_lot` feature is enabled

### Checklist

- [x] `Cargo.lock.min` updated with `cargo +nightly generate-lockfile -Z minimal-versions` + `cargo lock translate -v1`.
- [x] ~~CI job with minimal Rust version `1.36.0` for `parking_lot` feature added~~
- [x] `rustfmt`ed
- [x] `CHANGELOG.md` entry added


matklad:
While we are at it, it also makes sense to use `MaybeUninit` here: https://github.com/matklad/once_cell/blob/ea4634cef717e3d57b456460557426fc79cd4cbb/src/imp_pl.rs#L12. And, as this bumps MSRV, it should bump to 1.4.0.

.travis.yml
```diff
@@ -36,3 +36,8 @@ matrix:
     script:
     - mv Cargo.lock.min Cargo.lock
     - cargo build
+
+  - rust: 1.36.0
```
matklad: I think I am ok with only checking `parking_lot` with most recent stable.

tyranron:
@matklad 

> While we are at it, it also makes sense to use `MaybeUninit` here

Maybe as a separate PR?

matklad:
It makes sense to do this in the same PR where we bump MSRV, because that's exactly wha allows us ot use maybe uninit

tyranron:
@matklad also, what is the proper way for you to update `Cargo.lock.min` to V1 lock-file format?

matklad:
Don't know about v1, haven't looked into it yet. In general, it's best to just generate .min.lock using Cargo 1.31

tyranron:
Hmm... I've used `cargo +nightly generate-lockfile -Z minimal-versions` + `cargo lock translate -v1` and it seems to work well.

tyranron:
@matklad I've landed the initial implementation via `MaybeUninit` for `sync::OnceCell`. This is majorly based on #72 (thank you, @pitdicker!).

As I have a little experience dealing with `unsafe`s and low-level stuff like uninitialized memory, please make an early review of this. Once it's OK, I've land the similar changes for `Lazy`.

@pitdicker I'll be happy to see your comments too!

src/lib.rs
```diff
@@ -674,8 +674,15 @@ pub mod sync {
...
         /// Returns `None` if the cell is empty.
         pub fn get_mut(&mut self) -> Option<&mut T> {
-            // Safe b/c we have a unique access.
-            unsafe { &mut *self.0.value.get() }.as_mut()
+            if self.0.is_initialized() {
+                // Safe b/c we have a unique access.
+                unsafe {
+                    let slot: &mut MaybeUninit<T> = &mut *self.0.value.get();
+                    Some(&mut *slot.as_mut_ptr())
+                }
```
WaffleLapkin: It's safe not only because of the uniqueness of the access, but also b/c of the initialized flag. Not sure if it's obvious enough to not mention here.

src/lib.rs
```diff
@@ -822,10 +822,44 @@ pub mod sync {
...
+            // Safe b/c we have a unique access.
+            unsafe { self.0.mark_uninitialized() };
+
+            value
+        }
+
+        /// Takes the wrapped value out of a `OnceCell`.
+        /// Afterwards the cell is no longer initialized, and must be free'd
+        /// WITHOUT dropping. Only used by `into_inner` and `drop`.
+        fn take_inner(&mut self) -> Option<T> {
```
WaffleLapkin: This function takes out the value of the cell without changing initialized flag, this could lead to double-free even in simple code like this:
```rust
let cell = OnceCell::new();
cell.set(vec![0]);
let inner = cell.take_inner();
// drop inner
// drop cell => drop inner again
```

Несмотря на то что эта функция приватная, она всё равно должна быть помечана unsafe

WaffleLapkin: This function takes out the value of the cell without changing initialized flag, this could lead to double-free even in simple code like this:
```rust
let cell = OnceCell::new();
cell.set(vec![0]);
let inner = cell.take_inner();
// drop inner
// drop cell => drop inner again
```

Despite the fact that this is private API, this function must be `unsafe`.

src/imp_pl.rs
```diff
@@ -56,13 +57,30 @@ impl<T> OnceCell<T> {
...
+    /// It's not intended for concurrent use, that's why using
+    /// `Ordering::Relaxed` and accepting `&self` made for performance here.
+    ///
+    /// It's up to the caller to guarantee that:
+    /// - this function is not called concurrently (should be used where
+    ///   `&mut self` or `self` is accepted);
+    /// - the value contained in this cell is properly moved out before calling
+    ///   this function.
+    ///
+    /// Violating the rules above will cause UB (undefined behavior).
```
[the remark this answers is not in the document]
tyranron: @WaffleLapkin

> so what's the point of accepting `&self` (instead `&mut self`)

I've thought about possible better compiler optimizations.

> And the second case doesn't seem to be UB at all. If you think that you have uninitialized value, but actually you have an initialized value, then it could lead to memory leak, but not UB.

But what about overwritten value not being properly dropped?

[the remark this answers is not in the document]
WaffleLapkin: > I've thought about possible better compiler optimizations.

But with `&mut` compiler knows more and so could optimize better, doesn't it?

> But what about overwritten value not being properly dropped?

Forgetting to drop is not **UB**, it's only (a possible) memory leak. See [mem::forget](https://doc.rust-lang.org/std/mem/fn.forget.html)

src/imp_pl.rs
```diff
@@ -56,13 +57,30 @@ impl<T> OnceCell<T> {
             // - finally, if it returns Ok, we store the value and store the flag with
             //   `Release`, which synchronizes with `Acquire`s.
             let value = f()?;
-            let slot: &mut Option<T> = unsafe { &mut *self.value.get() };
-            debug_assert!(slot.is_none());
-            *slot = Some(value);
+            unsafe { self.value.get().write(MaybeUninit::new(value)) };
```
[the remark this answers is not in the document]
matklad: There's already a safety comment explanation on the function. We can't be more fine grained, as the `let _guard` above is what actully guarantees safety .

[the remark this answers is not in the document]
matklad: I'd keep the overall structure here:

```
let slot: &mut MaybeUninit<T> = unsafe { &mut *self.value.get() };
slot.as_mut_ptr().write(value);
```

src/imp_std.rs
```diff
@@ -6,22 +6,18 @@
 use std::{
     cell::{Cell, UnsafeCell},
     marker::PhantomData,
+    mem::MaybeUninit,
```
matklad: THe std impl should not change, because we need to maintain 1.31 MSRV with only std feature enabled

src/imp_pl.rs
```diff
@@ -9,7 +10,7 @@ use parking_lot::{lock_api::RawMutex as _RawMutex, RawMutex};
 pub(crate) struct OnceCell<T> {
     mutex: Mutex,
     is_initialized: AtomicBool,
-    pub(crate) value: UnsafeCell<Option<T>>,
+    pub(crate) value: UnsafeCell<MaybeUninit<T>>,
```
matklad: Ah, I've forgot that this is `pub(crate)`, this'll make this refactoring more annoying. Yeah, I guess you were right when you've suggested a separate PR, but we can still go either way. 

To do this cleanly we should first prepare code for refactoring in one commit, and then do refactoring in the second commeit. 

The first phase would be to make `value` here and in std `private`, and instead provide the required accessors (something like `fn as_ptr(&self) -> *mut T`). 

The second phase than should change to `MaybeUninit` only inside this module, without touching anything else outside.

src/imp_std.rs
tyranron: @matklad why we need to maintain 1.31 for `std`? It's a principal position on MSRV or any other arguments are involved?

src/imp_pl.rs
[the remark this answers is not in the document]
tyranron: @WaffleLapkin makes sense. Thanks for clarifying this!

src/imp_std.rs
matklad: Two things:

*First,* it is my personal position that foundational utility crates should be pretty conservative with MSRV, where for me "pretty conservative" is something like 10-ish Rust releases.

*Second,* we've already [documented](https://docs.rs/once_cell/1.3.1/once_cell/#minimum-supported-rustc-version) that the strategy for this crate is conservative, and we can't back it out.

tyranron:
@matklad what we have now:
- `MaybeUninit` is used only for `parking_lot` feature;
- MSRV is increased only when `parking_lot` feature is enabled;
- `get_unchecked`, `get_mut` and `into_inner` accessors implementations moved into submodules (it turned out the most reasonable way to avoid unnecessary casts and inconsistencies);
- all branch changes were splitted to separate meaningful commits.

src/imp_pl.rs
```diff
@@ -74,32 +74,76 @@ impl<T> OnceCell<T> {
...
         // verifies that it is not currently borrowed.
         // So, it is safe to move out `Option<T>`.
-        self.value.into_inner()
+        //
+        // Safe b/c marking this `OnceCell` as uninitialized below.
+        let value = unsafe { self.take_inner() };
+        if value.is_some() {
+            // `Relaxed` is OK here, because `self` is taking by value, so no
+            // concurrent operations are possible.
+            self.is_initialized.store(false, Ordering::Relaxed);
```
matklad: Rather than defusing the drop by pretending that the cell is uninitialized, I would prefer to do something like this:

```rust
pub(crate) fn into_inner(mut self) -> Option<T> {
    if !self.is_initialized() { return None; }
    let slot: &MaybeUninit<T> = &*self.value.get();
    let value: T = unsafe { ptr::read(slot.as_ptr()) };
    mem::forget(self);
}

impl<T> Drop for OnceCell<T> {
    fn drop(&mut self) {
        if !self.is_initialized() { return; }
        let slot: &mut MaybeUninit<T> = &mut *self.value.get();
        unsafe { mem::drop_in_place(slot.as_mut_ptr()) };
    }
}
```

It might also make sense to add a couple of private helpers to stop repeating the slot thing everyhere

```rust
impl OnceCell<T> {
    fn as_ptr(&self) -> *const T;
    fn as_ptr_mut(&mut self) -> *mut T;
}
```

matklad:
bors r+

Let's land this now! I think there's a couple of tweaks I want to do, but I've also found an unrelated obscure bug which I'd love to fix, so I do the changes myself

PR #99, opened 2020-04-07
miri: enable leakcheck

RalfJung:
https://github.com/rust-lang/miri/issues/940 is fixed, so we should be able to enable the leak checker now. :)

(However, we'll have to wait until https://github.com/rust-lang/rust/pull/70897 lands so that this is available via rustup.)


RalfJung:
A nightly that that PR got published, cycling PR to re-trigger CI.

RalfJung:
(CI is green on Travis, not sure why it does not report back here.)

matklad:
Perfect, thanks!

(as an aside, I plan to put some implmentation effort into moving this crate into stdlib soon)

PR #100, opened 2020-05-04
add a `take` method which takes a mutable reference

sunshowers:
I use `OnceCell` extensively in my code to cache lazily computed data on
structs. Sometimes, I mutate the data -- in which case I have to invalidate
those caches.

This is of course possible to do by oneself, but it's slightly non-obvious
that the "once" bit of OnceCell only applies in the shared reference context.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -246,7 +246,7 @@ mod imp;
 pub mod unsync {
     use core::{
         cell::{Cell, UnsafeCell},
-        fmt,
+        fmt, mem,
         ops::{Deref, DerefMut},
     };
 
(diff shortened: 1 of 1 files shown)


matklad:
Hm, I am a bit on the fence about this... My first reaction is that `mem::take` should be good enough. My second thought is that [`Option::take`](https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.take) is a thing....

What do you think about renaming `reset` to `take`?

sunshowers:
Sure, sounds reasonable to me for symmetry.

sunshowers:
done, thanks!

sunshowers:
ah, looks like `mem::take` isn't on the MSRV 1.31.1. think the MSRV should be bumped to 1.40?

matklad:
I'd prefer not to bump MSRV unless necessary (really, I'd prefer to not bump MSRV at all, and just implement https://github.com/rust-lang/rfcs/pull/2788). `mem::take` is just `mem::replace(_, Default::default)`, so should not be hard to poly-fill.

matklad:
I'll fix this up myself, thanks for sending a PR and sorry for this taking so long to review :)

matklad:
Also, wow, this is the hundredth issue/pr... Didn't expect that from a relatively simple utility crate :)

matklad:
Published as 1.4.0: https://crates.io/crates/once_cell/1.4.0

PR #104, opened 2020-05-16
Implement serde support

Techcable:


matklad:
Hm, in my mind serde support doesn't belong to OnceCell. There's an impedencse mismatch here between data at rest, which is just data, and lazy data in once cell. The caller should decide how to fix this mismatch, not the library.

Techcable:
For my use case, I'm using OnceCell to defer the computation of a value in a cache. At times I serialize that whole cache to disk for future process runs (since my computation is pure). I feel like this is a major use case for an `OnceCell`. The [cached crate](https://crates.io/crates/cached) uses OnceCell to cache pure functions as well. It has about 74k downloads.

Serializing the OnceCell as an `Optional` is the obvious default for a cache. I don't really think there's another serialization model that would make sense, regardless of how OnceCell is used.

You're right that Serialization is awkward in many cases (especially when the lazy data is related to application state), but it''s reasonable to make it a builtin option.
Serde support would be off by default, so users would have to think through the appropriateness of serialization.

matklad:
I thought about this some more, and decided that serialization is out of scope for this crate.

Users should do either of:

* using serde's `serialize_with` / `deserialize_with` hooks
* write a custom wrapper around `OnceCell`
* use some other crate from crates.io which implements OnceCell+Serialization.

PR #106, opened 2020-05-27
Add an into_inner function to Lazy

ninegua:
This is useful to avoid an extra clone when the purpose is to take out the Lazy value.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -954,6 +954,22 @@ pub mod sync {
                 None => panic!("Lazy instance has previously been poisoned"),
             })
         }
+
+        /// Forces the evaluation of this lazy value and
+        /// returns the result.
+        ///
+        /// # Example
+        /// ```
+        /// use once_cell::sync::Lazy;
+        ///
+        /// let lazy = Lazy::new(|| 92);
+        ///
+        /// assert_eq!(lazy.into_inner(), 92);
+        /// ```
+        pub fn into_inner(self) -> T {
+            Lazy::force(&self);
+            self.cell.into_inner().unwrap()
+        }
     }
 
     impl<T, F: FnOnce() -> T> Deref for Lazy<T, F> {


src/lib.rs
```diff
@@ -954,6 +954,22 @@ pub mod sync {
...
+        /// use once_cell::sync::Lazy;
+        ///
+        /// let lazy = Lazy::new(|| 92);
+        ///
+        /// assert_eq!(lazy.into_inner(), 92);
+        /// ```
+        pub fn into_inner(self) -> T {
+            Lazy::force(&self);
+            self.cell.into_inner().unwrap()
+        }
```
matklad, line 972: I feel that forcing and destructing should be orthogonal, so I don't feel good about a function that does both. My gut feeling would be that the signature should look like `fn into_inner(self) -> Option<T>`, but then we loose the `F`...

ninegua: Returning `Option<T>` will give away the information of whether the callback has been called or not, which is something that the current `Lazy` interface does not provide. I kind of like it that way, which is why I chose to `force` the value before returning it. If people feel strongly about introspection, maybe they should use `OnceCell` instead of `Lazy`?

matklad: I see now problem adding `Lazy::is_empty(this: &Self) -> bool` method, it's just that nobody asked for that. 

In general, I feel that APIs for data-strcuture things should not be prescriptive, and should err on the side of generality. Basically, let the user decide how to use the thing.

ninegua: What is the plan of moving forward?

If we introduce a destructor with type `Lazy<T> -> Result<T, FnOnce() -> T>`, it likely should go under a different name than `into_inner`, which by convention should have type `Lazy<T> -> T`. Maybe call it `extract`?

In any case, I'm not opposed to adding `extract` or `is_empty` to the interface, and I don't think they are mutually exclusive to an `into_inner` function either. Maybe they can go into a separate PR. What do you think? @matklad

camsteffen: How about naming this method `into_value`? It seems useful as is, just not appropriately named.

matklad: Hm, I still think that neither name nor signature are optimal. Let's dodge the issue by introducing a roughly equivalent API, which does have a perfect name:

```
impl Lazy<T> {
  fn take(&mut self) -> Option<T> { ... }
}
```

camsteffen: I'm surprised that `take` was added to `OnceCell`. I thought a central tenant of `OnceCell` is "can be assigned to at most once" (at the top of the readme). Isn't that the meaning of the name of the struct? Adding `take` to `Lazy` would remove the (I think important) semantic that the initializer will run at most once.

A positive of having a function that does both forcing and destructing is that the synchronization can (at least potentially) be optimized as a single operation. To offer one more name idea: `force_into_value`.

sunshowers: (I'm the one that added take.)

once_cell's at-most-once guarantee applies in the shared reference context. If you have a mutable reference you can do whatever with it. (I had a comment explaining this but @matklad removed it I think ;) )

camsteffen: That makes sense.

ninegua: I think the functionality of `into_inner` here is not merely a convenience. Yes, now with `take` we could implement a similar thing, but it is awkward to have an unnecessary unwrap in there.

camsteffen: Right. Perhaps there could be `force_take(&mut self) -> T`. I don't mean to flippantly keep adding ideas. I'm hoping something will stick that we can agree on.

matklad:
Thought a little bit more about this. take-stype API is impossible, but we can use a `Result` to deconstruct lazy into value/fn "pair": https://github.com/matklad/once_cell/pull/131

matklad:
closed by https://github.com/matklad/once_cell/pull/131

PR #112, opened 2020-08-19
enable some thread-using tests on Miri

RalfJung:


matklad:
Ah, missed those, thanks!

bors r+

And, as you are going to be bugged by a notification anyway, you might want to look at completely unrelated https://github.com/matklad/once_cell/pull/110/commits/d5295146e2fa80f4e7ad455fd97d21e3002ec885 which exposed an interesting API quirk of `MaybeUninit`. If you use `MaybeUninit`, you have to write your `Drop` impl, which requires an (unstable) `may_dangle` to be dropck transparent. Am I correct that there's no way to simulate `ManyallyDrop + may_dangle` on stable, other than just  using `Option` instead of `MaybeUninit`

RalfJung:
Presumably you need the `Drop` impl because the field is not always initialized when the container is dropped -- so even if the UB rules were different and `MaybeUninit` was not necessary, that still wouldn't help, it would just be wrong to always drop?

Yeah that is an interesting quirk that I have not seen yet. Might be worth bringing up in the rust-lang/rust repo in case it hasn't already.

PR #114, opened 2020-09-28
Fix alignment of `imp_std::Waiter`

eduardosm:
Using `#[repr(align(4))]` directly on `Waiter` will downgrade alignment form 8 to 4 bytes on 64-bit targets.

To fix this, an empty struct is defined (`Align4`), which is tagged with `#[repr(align(4))]`, and an instance of this empty struct is added to `Waiter`. `Waiter` will now have an alignment of at least 4 bytes, but it will be greater on targets that require it.

### DIFF
--- a/src/imp_std.rs
+++ b/src/imp_std.rs
@@ -46,12 +46,17 @@ const COMPLETE: usize = 0x2;
 // this is in the RUNNING state.
 const STATE_MASK: usize = 0x3;
 
+// Empty struct that is 4-byte aligned.
+#[repr(align(4))]
+struct Align4;
+
 // Representation of a node in the linked list of waiters in the RUNNING state.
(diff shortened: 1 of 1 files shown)


matklad:
Hm, I don't think this how `align` works?

>For align, if the specified alignment is less than the alignment of the type without the align modifier, then the alignment is unaffected.

from https://doc.rust-lang.org/reference/type-layout.html#the-alignment-modifiers

eduardosm:
Ouch, I mixed the `align` and `packed` concepts, sorry. Closing this PR then.

matklad:
It's all good!

JFYI, here's how I figured that align might not work that way:

If align were to force the alignment, than it would be unsafe to do field access with such a struct. Given that there's no unsafe in the attribute, and the compiler doesn't require unsafe when accessing fields (which would be the case for packed), then it seems likely that it only sets the min alighment.

PR #137, opened 2021-02-24
Convert race to a default-enabled feature for targets without atomics.

### LINKED ISSUE #136: `race` module cannot be used on targets without atomics
As of `v1.6.0`, the `race` module became stable. I've been using `once_cell` in my msp430 code since `v1.2.0` without issue. Unfortunately, since the release of `v1.6.0`, I've been unable to compile the module:

```
William@DESKTOP-H0PMN4M MINGW64 ~/Projects/rust-embedded/once_cell
$ cargo build --release -Zbuild-std=core --target=msp430-none-elf --no-default-features
   Compiling once_cell v1.6.0 (C:\msys64\home\William\Projects\rust-embedded\once_cell)
error[E0432]: unresolved import `core::sync::atomic::AtomicUsize`
  --> src\race.rs:11:20
   |
11 |     sync::atomic::{AtomicUsize, Ordering},
   |                    ^^^^^^^^^^^ no `AtomicUsize` in `sync::atomic`

error[E0283]: type annotations needed
  --> src\race.rs:17:5
   |
17 |     inner: AtomicUsize,
   |     ^^^^^^^^^^^^^^^^^^ cannot infer type
   |
   = note: cannot satisfy `_: Default`
   = note: required by `core::default::Default::default`
   = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)

error: aborting due to 2 previous errors

Some errors have detailed explanations: E0283, E0432.
For more information about an error, try `rustc --explain E0283`.
error: could not compile `once_cell`

To learn more, run the command again with --verbose.
```

Is it possible to gate the `race` module behind a feature so than I can continue to use `once_cell` on a target without atomics, or would this be considered a semver breaking change?

cr1901:
Closes #136.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,9 @@ default = ["std"]
 # Enables `once_cell::sync` module.
 std = ["alloc"]
 # Enables `once_cell::race::OnceBox` type.
-alloc = []
+alloc = ["race"]
+# Enables `once_cell::race` module.
+race = []
 # Enables semver-exempt APIs of this crate.
 # At the moment, this feature is unused.
 unstable = []
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1065,4 +1065,5 @@ pub mod sync {
     fn _dummy() {}
 }
 
+#[cfg(feature = "race")]
 pub mod race;
--- a/tests/it.rs
+++ b/tests/it.rs
@@ -589,7 +589,7 @@ mod sync {
     }
 }
 
-#[cfg(feature = "unstable")]
+#[cfg(all(feature = "unstable", feature = "race"))]
 mod race {
     use std::{
         num::NonZeroUsize,


Cargo.toml
```diff
@@ -34,7 +34,9 @@ default = ["std"]
 # Enables `once_cell::sync` module.
 std = ["alloc"]
 # Enables `once_cell::race::OnceBox` type.
-alloc = []
+alloc = ["race"]
+# Enables `once_cell::race` module.
+race = []
```
matklad, line 39: This should be added to `default` above I think?

cr1901: Oh, it's not transitive (`std` depends on `alloc` with depends on `race`)? Hmmm... anyway CI fails, so I better fix that :P.

matklad: Ah, you are of course right!

cr1901:
CI fails in [PUBLISH](https://github.com/matklad/once_cell/runs/1971232614#step:3:995)... I don't think there's much I can do about that?

matklad:
(re-triggering the CI, the error should be fixed by https://github.com/matklad/xaction/commit/c29784438605b983d0d60fc7c21e27aa2374cefe)

PR #140, opened 2021-03-02
Improve `imp_pl` size

quininer:
I have observed that the program size has increased a lot after enabling the `parking_lot` feature.

`parking_lot` will actively inline methods, which leads to some unnecessary duplication of code. see https://github.com/Amanieu/parking_lot/pull/236

This PR introduces an `InlineLessRawMutex` wrapper to avoid this problem. `initialize` is a cold path, I don't think this will cause performance problems.

### DIFF
--- a/src/imp_pl.rs
+++ b/src/imp_pl.rs
@@ -5,7 +5,7 @@ use std::{
     sync::atomic::{AtomicBool, Ordering},
 };
 
-use parking_lot::lock_api;
+use parking_lot::Mutex;
 
 pub(crate) struct OnceCell<T> {
     mutex: Mutex<()>,
@@ -27,7 +27,7 @@ impl<T: UnwindSafe> UnwindSafe for OnceCell<T> {}
 impl<T> OnceCell<T> {
     pub(crate) const fn new() -> OnceCell<T> {
         OnceCell {
-            mutex: const_mutex(()),
+            mutex: parking_lot::const_mutex(()),
             is_initialized: AtomicBool::new(false),
             value: UnsafeCell::new(None),
         }
@@ -46,8 +46,10 @@ impl<T> OnceCell<T> {
     where
         F: FnOnce() -> Result<T, E>,
     {
-        let _guard = self.mutex.lock();
-        if !self.is_initialized() {
+        let mut f = Some(f);
+        let mut res: Result<(), E> = Ok(());
+        let slot: *mut Option<T> = self.value.get();
+        initialize_inner(&self.mutex, &self.is_initialized, &mut || {
             // We are calling user-supplied function and need to be careful.
             // - if it returns Err, we unlock mutex and return without touching anything
             // - if it panics, we unlock mutex and propagate panic without touching anything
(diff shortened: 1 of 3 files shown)


matklad:
I think if the goal here is solely to reduce code size due to PL, than the problem should be either solved upstream, or worked-around downstream. Placing a workaround into an intermediate library isnt’ worth it, imo :P

Curious, would slappping `inline(never)` onto `initialize` help?


We might also want to ditch pl mutexes altogether, and use the underlying parking mechanism directly. That should also reduce the size of cell itself, I think. I’d be willing to accept/write a patch for that, but it should be written in a very clean style, such that it’s easy to see that it is correct.

quininer:
The parking_lot cannot determine whether this is a cold path, so putting this method upstream is not appropriate.

> Curious, would slappping inline(never) onto initialize help?

This won't help, because initialize is a generic method.

matklad:
> This won't help, because initialize is a generic method.

Ag, *that* is a bug I think! Take a look at the std impl for the trick to un-generalize it.

quininer:
I added `initialize_inner` with reference to `imp_std`.

src/imp_pl.rs
```diff
@@ -56,15 +58,22 @@ impl<T> OnceCell<T> {
...
-            let value = f()?;
-            // Safe b/c we have a unique access and no panic may happen
-            // until the cell is marked as initialized.
-            let slot: &mut Option<T> = unsafe { &mut *self.value.get() };
-            debug_assert!(slot.is_none());
-            *slot = Some(value);
-            self.is_initialized.store(true, Ordering::Release);
-        }
-        Ok(())
+            let f = f.take().unwrap_or_else(|| unsafe { hint::unreachable_unchecked() });
```
matklad, line 61: This probably shaves off a couple of byets, right? Let's exctract this to  an `unwrap_unchecked` function into the parent module, and use it for `impl_std` as well.

src/imp_pl.rs
```diff
@@ -102,6 +111,17 @@ impl<T> OnceCell<T> {
     }
 }
 
+#[inline(never)]
+fn initialize_inner(mutex: &Mutex<()>, is_initialized: &AtomicBool, init: &mut dyn FnMut() -> bool) {
```
matklad, line 115:
```suggestion
#[inline(never)]
// Note: this is intentionally monomorphic
fn initialize_inner(mutex: &Mutex<()>, is_initialized: &AtomicBool, init: &mut dyn FnMut() -> bool) {
```

matklad (commented):
LGTM! could you also add a changelog entry (after rebasing on master, ther's another PR in flight) and bump a patch in Cargo.toml?

src/imp_pl.rs
quininer: `unwrap_or_else` is short enough, I don't think it needs to be extracted into a separate function.

matklad: fine by me, as the two code paths for pl and for std are identical

matklad:
bors r+

Thanks!

PR #141, opened 2021-03-02
Fix race::OnceBox<T>: Default to not require T: Default

Arnavion:


matklad:
d'oh, could you add a changelog entry and bump a patch version in Cargo.toml?

Arnavion:
Done.

PR #148, opened 2021-03-16
Return value from `set`

daxpedda:
This adds a function called `set_and_get` that returns the value on setting it.
Originally I wanted to implement it only for `sync`, because the value was just being discarded, see the comment I added to `unsync::OnceCell::set_and_get`.

Is there interest in a feature like this?

EDIT:
My use case is to initialize but return an `Err` if it already is initialized. This can be done by calling `OnceCell::set` and then `OnceCell::get`, this offers an optimization, that culls this to only one function call.
This can't be solved by using `OnceCell::get_or_init` because it doesn't offer any indication if it was already set.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -476,6 +476,37 @@ pub mod unsync {
             Ok(())
         }
 
+        /// Sets the contents of this cell to `value` and returns a reference to it.
+        ///
+        /// Returns `Ok(&value)` if the cell was empty and `Err(value)` if it was
+        /// full.
+        ///
+        /// # Example
+        /// ```
+        /// use once_cell::unsync::OnceCell;
+        ///
+        /// let cell = OnceCell::new();
+        /// assert!(cell.get().is_none());
+        ///
+        /// assert_eq!(cell.set_and_get(92), Ok(&92));
+        /// assert_eq!(cell.set_and_get(62), Err(62));
+        ///
+        /// assert!(cell.get().is_some());
+        /// ```
+        pub fn set_and_get(&self, value: T) -> Result<&T, T> {
+            let slot = unsafe { &*self.inner.get() };
+            if slot.is_some() {
+                return Err(value);
+            }
+            let slot = unsafe { &mut *self.inner.get() };
+            // This is the only place where we set the slot, no races
+            // due to reentrancy/concurrency are possible, and we've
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -476,6 +476,37 @@ pub mod unsync {
...
+            if slot.is_some() {
+                return Err(value);
+            }
+            let slot = unsafe { &mut *self.inner.get() };
+            // This is the only place where we set the slot, no races
+            // due to reentrancy/concurrency are possible, and we've
+            // checked that slot is currently `None`, so this write
+            // maintains the `inner`'s invariant.
+            *slot = Some(value);
+            Ok(slot.as_ref().unwrap())
```
daxpedda, line 507: It's possible to use `unsafe` to extract the value here instead of using `unwrap`.

daxpedda: Like this basically:
```suggestion
            if let Some(value) = slot.as_ref() {
                Ok(value)
            } else {
                unsafe { core::hint::unreachable_unchecked() }
            }
```

matklad:
I am torn on this one. One the one hand, `set` not returning a reference to a value is definitelly a loss of generality. On the other hand, the fully general signature here would be `-> Result<&T, (&T, T)>`. On the third hand, it feels like `.set`, `.get` and `.get_or_init` give the necessary tools to express this without adding extra methods. 

Anothe option would be to add a simpler `fn insert(&self, value: T) -> &T` methond, a-la `Option<Insert>`.

daxpedda:
> On the other hand, the fully general signature here would be `-> Result<&T, (&T, T)>`. 

I don't mind implementing it like this if you want.
> On the third hand, it feels like `.set`, `.get` and `.get_or_init` give the necessary tools to express this without adding extra methods.

So the only reason `.get_or_init` doesn't cover this for me is because it doesn't tell me if it was already initialized. I want an error if it's already initialized. So I'm kinda forced to use `.get` and `.set` to achieve this.
> Anothe option would be to add a simpler `fn insert(&self, value: T) -> &T` methond, a-la `Option<Insert>`.

Same as above, it doesn't solve my use-case.

Sorry, I think I should have explained my use-case more clearly in the OP.

matklad:
should be covered by #151

PR #151, opened 2021-05-28
Add try insert

matklad:
bors r+
🤖


correabuscar:
I was surprised to see that `set(value)` returns the `value` that was attempted to be set in `Err(value)`, especially since the `try_insert` that's based upon can return a ref to the value that was already set.

PR #154, opened 2021-06-22
Document implications of `&mut self` methods on once-ness

### LINKED ISSUE #153: Understanding "once" in the presence of `take()`
I got confused when I came across the `take()` method. Specifically, the README says (emphasis mine):

> `OnceCell` might store arbitrary non-`Copy` types, **can be assigned to at most once** and provide direct access to the stored contents.

But with `take()`, it's easy to assign to a `OnceCell` an arbitrary number of times:

```rust
let mut cell: OnceCell<u32> = OnceCell::new();

for i in 0..1000 {
    cell.set(i).unwrap();
    assert_eq!(cell.take(), Some(i));
}
```

This left me confused about what exactly the core abstraction of this API is aiming to accomplish. I suspect this is probably just a documentation issue? But can you help me understand what "once-ness" means given that you can actually assign to the cell an arbitrary number of times?

dherman:
Document the allowed violation of the "once" invariant for `&mut self` methods.

Closes #153.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -441,6 +441,18 @@ pub mod unsync {
         /// Gets a mutable reference to the underlying value.
         ///
         /// Returns `None` if the cell is empty.
+        ///
+        /// This method is allowed to violate the invariant of writing to a `OnceCell`
+        /// at most once because it requires `&mut` access to `self`. As with all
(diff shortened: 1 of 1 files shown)


matklad:
Looks perfect, thanks!

bors r+

PR #157, opened 2021-08-08
Async support

danieldg:
This adds new types for thread-safe cells whose initialization functions are async.  This must be distinct from the sync cell because Futures-style blocking should not be used to block sync threads (and vice versa).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,9 +34,11 @@ default = ["std"]
 # Enables `once_cell::sync` module.
 std = ["alloc"]
 # Enables `once_cell::race::OnceBox` type.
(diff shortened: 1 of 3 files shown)


matklad:
I suggest publishing this as a separate crate (which can be linked-to from onec_cell docs): the code is very different from what once_cell is doing, so I don't think there's a lot of benefit in keeping this in the same crate.

danieldg:
Published as https://crates.io/crates/async-once-cell

matklad:
Took a while, but pointer to async_once_cell is now in the docs! https://docs.rs/once_cell/latest/once_cell/#faq

PR #165, opened 2021-12-01
Add an atomic-polyfill optional dependancy

### LINKED ISSUE #155: once_cell::race does not compile on ARM
These four lines: [1](https://github.com/matklad/once_cell/blob/master/src/race.rs#L41), [2](https://github.com/matklad/once_cell/blob/master/src/race.rs#L82), [3](https://github.com/matklad/once_cell/blob/master/src/race.rs#L214), [4](https://github.com/matklad/once_cell/blob/master/src/race.rs#L260) fail to compile on ARM because on ARM the `AtomicUsize` lacks a `compare_exchange` method.
It may be possible to work around this with a `load` and a `store` and a loop.

ogoffart:
To make it compile on platform that have limited atomics

Fixes #155 

Note that this re-implement the Debug version of the OnceBox
because atomic-polyfill::AtomicPtr don't implement Debug (https://github.com/embassy-rs/atomic-polyfill/issues/11)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.9
+
+- Added an `atomic-polyfill` optional dependency to compile `race` on platforms without atomics
+
 ## 1.8.0
 
 - Add `try_insert` API -- a version of `set` that returns a reference.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "once_cell"
-version = "1.8.0"
+version = "1.9.0"
 authors = ["Aleksey Kladov <aleksey.kladov@gmail.com>"]
 license = "MIT OR Apache-2.0"
 edition = "2018"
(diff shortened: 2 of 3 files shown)


ogoffart:
Ping?

ogoffart:
The use-case is indeed to be able to make cross platform libraries that use once_cell::race so they can be used on both platform with a std library, and also on bare metal embedded on device that do not have threads and atomic.

> I am not a huge fan of the solutions in the style of "silently replace mutexes with spinlocks" -- I generally thing the caller should be responsible for selecting the right synchronization primitive.

I have read https://matklad.github.io/2020/01/02/spinlocks-considered-harmful.html and I agree 100%
But in this case, there is probably no spinlock involved, even when this feature is enabled
If this is a platform that supports atomic, then the core::sync::atomic classes are used so no change at all.
If the platform does not support atomic, it depends on how the critical_section crates implements it. In single core scenario, this is just about disabling the interrupt for the few instructions. On multi core device, there might be a spinlock. but how else could we implement it?

> What happens if there are multiple cores/CPUs in play)? Or are we sure that that that's not a problem for these targets?

The atomic-polyfill README  states:  https://github.com/embassy-rs/atomic-polyfill

> > Note: polyfill is based on critical sections using the [`critical-section`](https://crates.io/crates/critical-section) crate. The default implementation is based on disabling all interrupts, so it's **unsound** on multi-core targets. It is possible to supply a custom 
critical section implementation, check the `critical-section` docs for details.

> Could you bump a version in Cargo.toml as well, such that this gets a release automatically when merged?

Should it be 1.8.1 or 1.9 ?

matklad:
I guess, my initial knee-jerk reaction was overly negative. I would push back strongly against this being enabled *by default*, but having sketchy things under a feature flag is totally fine :-)

matklad:
bors r+

Thanks!

PR #170, opened 2022-02-05
Add `OnceCell::with_value` and improve `Clone` implementation

### LINKED ISSUE #164: Feature Request: `OnceCell::with_value`
Creates a `OnceCell` that is already initialized with the provided value, like 
```rust
let x = OnceCell::new();
let _ = x.set(/* stuff */);
```

I have a use case where there's two paths to creating the `OnceCell`, one where the contents are already known, and one where they're calculated later

I could probably implement this, if that's cool

a1phyr:
This PR adds `OnceCell::with_value` to both `sync::OnceCell` and `unsync::OnceCell`. This enables creating an initialized `OnceCell` at compile time and/or without synchronization cost.

Additionally, this improve `Clone` implementations by removing a panicking branch, reducing synchronization cost and overriding `clone_from` implementation.

Closes #164

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 1.11
+
+- Add `OnceCell::with_value` to create initialized `OnceCell` at compile time.
+- Improve `Clone` implementation for `OnceCell`.
+
 ## 1.10
 
 - upgrade `parking_lot` to `0.12.0` (note that this bumps MSRV with `parking_lot` feature enabled to `1.49.0`).
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ atomic-polyfill = { version = "0.1", optional = true }
 
 [dev-dependencies]
 lazy_static = "1.0.0"
-crossbeam-utils = "0.7.2"
+crossbeam-utils = "0.8.7"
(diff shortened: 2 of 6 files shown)


matklad:
Hm, I am a bit torn here -- it feels like `From` is probably the best interface here. But that can't be const. On the other hand, const-creating a full OnceCell feels a bit odd. Do we have a real use-case for that? At the same time, the API does feel rather tidy. 

:thinking: yeah, I guess, lets have this!  The implementation looks good to me, but lets add a couple of tests to https://github.com/matklad/once_cell/blob/master/tests/it.rs to make sure the new API doesn't get broken by accident.

a1phyr:
Actually, I didn't see the `From` implementation when I wrote this patch, so I wrote `with_value` for that, so I don't have a strong will to add this.

There *might* be a use-case for something like that, I guess (though very specific):

```rust
const fn get_some_value() -> Option<T> {
    /* ... */
}

static CELL: OnceCell<T> = match get_some_value() {
    Some(value) => OnceCell::with_value(value),
    None => OnceCell::new(),
};
```

I think the main benefit of this would be discoverability, which is not great today, as shown by the linked issue (and my own experience ^^').

matklad (commented):
bors r+

Thanks!

PR #175, opened 2022-03-17
Explain the atomic orderings used in race

Kestrer:
I believe this is the correct rationale as to why those atomic orderings were chosen in this module. This change also makes it explicit that those orderings are used in the library's public API, better allowing consumers to safely depend on it.

### DIFF
--- a/src/race.rs
+++ b/src/race.rs
@@ -10,13 +10,14 @@
 //!
 //! All types in this module use `Acquire` and `Release`
 //! [atomic orderings](Ordering) for all their operations. While this is not
-//! strictly necessary, it is useful for users as it allows them to be certain
-//! that after `get` or `get_or_init` returns on one thread, any side-effects
-//! caused by the setter thread prior to them calling `set` or `get_or_init`
-//! will be made visible to that thread; without it, it's possible for it to
-//! appear as if they haven't happened yet from the getter thread's perspective.
-//! This is an acceptable tradeoff to make since `Acquire` and `Release` have
-//! very little performance overhead on most architectures versus `Relaxed`.
+//! strictly necessary for types other than `OnceBox`, it is useful for users as
+//! it allows them to be certain that after `get` or `get_or_init` returns on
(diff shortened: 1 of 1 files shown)


src/race.rs
```diff
@@ -5,6 +5,18 @@
 //! them stores the result.
 //!
 //! This module does not require `std` feature.
+//!
+//! # Atomic orderings
+//!
+//! All types in this module use `Acquire` and `Release`
+//! [atomic orderings](Ordering) for all their operations. While this is not
+//! strictly necessary, it is useful for users as it allows them to be certain
```
matklad, line 13: This actually **is** strictly necessary for `OnceBox` I believe. Othewise, you could leak partialy constructed objects

matklad:
Good call! We indeed should document acq/rel guarantee!

src/race.rs
Kestrer: Yeah, I think you might be right. I'll update it once I have access to a computer.

src/race.rs
```diff
@@ -10,13 +10,14 @@
...
-//! This is an acceptable tradeoff to make since `Acquire` and `Release` have
-//! very little performance overhead on most architectures versus `Relaxed`.
+//! strictly necessary for types other than `OnceBox`, it is useful for users as
+//! it allows them to be certain that after `get` or `get_or_init` returns on
+//! one thread, any side-effects caused by the setter thread prior to them
+//! calling `set` or `get_or_init` will be made visible to that thread; without
+//! it, it's possible for it to appear as if they haven't happened yet from the
+//! getter thread's perspective. This is an acceptable tradeoff to make since
+//! `Acquire` and `Release` have very little performance overhead on most
+//! architectures versus `Relaxed`.
```
matklad, line 20: LGTM!

Bonus points if you can add a "litmus test" example here!

matklad (approved):

matklad:
bors d+

Issue #102, opened 2020-05-13
Add OnceCell::wait


matklad:
This seems plausible:

```rust
impl OnceCell<T> {
  /// Blocks until the cell is populated
  fn wait(&self) -> &T {}
}
```

The implementaiton should basically stuff the current thread into the waiters list, without attempting to set the value.

That3Percent:
For what it's worth, my opinion is that the addition of this API is a much better idea than making `get` block and #92 should not be merged.

matklad:
Hm, it's interesting that this is slightly not-trivial. Today, when a thread does `get_or_try_init` and fails (ie, it does not actually initialize the cell), it wakes all other waiters. This makes sense, as every waiter can set the cell.

With `wait`, we'll get waiters which ideally should not be woken up if we know that the cell is uninint.

PR #177, opened 2022-05-18
add OnceCell::wait

matklad:
issue: #102

### DIFF
--- a/src/imp_pl.rs
+++ b/src/imp_pl.rs
@@ -77,6 +77,21 @@ impl<T> OnceCell<T> {
         res
     }
 
+    #[cold]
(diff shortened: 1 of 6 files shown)


src/imp_pl.rs
```diff
@@ -86,6 +86,10 @@ impl<T> OnceCell<T> {
         res
     }
 
+    pub(crate) fn wait(&self) {
+        unimplemented!()
+    }
```
matklad, line 91: Does anyone know if it's actually possible to implement this with parking lot without incerasing the size_of onec_cell?

matklad:
bors r!

PR #178, opened 2022-05-18
replace parking_lot with parking_lot_core

matklad:
This is primaraliy to enable https://github.com/matklad/once_cell/pull/177 for pl as well

### DIFF
--- a/src/imp_pl.rs
+++ b/src/imp_pl.rs
@@ -2,18 +2,17 @@ use std::{
     cell::UnsafeCell,
     hint,
     panic::{RefUnwindSafe, UnwindSafe},
-    sync::atomic::{AtomicBool, Ordering},
+    sync::atomic::{AtomicU8, Ordering},
 };
 
-use parking_lot::Mutex;
-
-use crate::take_unchecked;
-
 pub(crate) struct OnceCell<T> {
-    mutex: Mutex<()>,
-    is_initialized: AtomicBool,
+    state: AtomicU8,
     value: UnsafeCell<Option<T>>,
 }
+
+const INCOMPLETE: u8 = 0x0;
+const RUNNING: u8 = 0x1;
+const COMPLETE: u8 = 0x2;
 
 // Why do we need `T: Send`?
 // Thread A creates a `OnceCell` and shares it with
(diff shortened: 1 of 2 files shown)


matklad:
@pitdicker if by any chance you have resources and desire to make a review here, that'd be helpful! No worries if you don't!

src/imp_pl.rs
```diff
@@ -28,25 +27,17 @@ impl<T: UnwindSafe> UnwindSafe for OnceCell<T> {}
...
-            value: UnsafeCell::new(Some(value)),
-        }
+        OnceCell { state: AtomicU8::new(COMPLETE), value: UnsafeCell::new(Some(value)) }
     }
 
     /// Safety: synchronizes with store to value via Release/Acquire.
     #[inline]
     pub(crate) fn is_initialized(&self) -> bool {
-        self.is_initialized.load(Ordering::Acquire)
+        self.state.load(Ordering::Acquire) == COMPLETE
```
nagisa, line 40: This can _probably_ be a plain `Relaxed` load.

The use of this API can only ever be used in an informational manner (it is 100% racy no matter how the returned value is used…) so it doesn't matter if it spuriously returns false for a little while when the once cell is already initialized.

src/imp_pl.rs
```diff
@@ -121,18 +112,47 @@ impl<T> OnceCell<T> {
...
-    let _guard = mutex.lock();
-
-    if !is_initialized.load(Ordering::Acquire) {
-        if init() {
-            is_initialized.store(true, Ordering::Release);
+fn initialize_inner(state: &AtomicU8, init: &mut dyn FnMut() -> bool) {
+    let key = state as *const _ as usize;
+    loop {
+        let exchange =
+            state.compare_exchange(INCOMPLETE, RUNNING, Ordering::Acquire, Ordering::Acquire);
```
nagisa, line 132:
```suggestion
            state.compare_exchange(INCOMPLETE, RUNNING, Ordering::Acquire, Ordering::Relaxed);
```

You could also consider `compare_exchange_weak`, though I don't know if there's any benefit in doing so since you will need to `continue` on `Err(INCOMPLETE)` anyway.

nagisa (commented):
Did you consider setting up `loupe` to actually verify the atomics?

src/imp_pl.rs
matklad: This one I am rather sure *should* be `Acquire`, as it guards `get_unchecked` accesses to the value. 

https://github.com/matklad/once_cell/blob/b57c77415b1146859402dcb905850fe9f44b7bb6/src/lib.rs#L864-L867

src/imp_pl.rs
matklad: Hm, this I think also needs to be acquire, otherwise `Err(COMPLETE) => return,` branch won't synchronize-with the corresponding Release store, and subsequent `get_unchecked` would read garbage. Both cases are great to check if loom works though!

PR #180, opened 2022-05-30
Fix comments

muramasa8191:
# Fix comments

The example of `get_mut` looks like that for `new`.
Actually,  the comment before the example finishes with ':'. So, it looks like in the middle of sentence or incomplete.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -434,6 +434,15 @@ pub mod unsync {
 
     impl<T> OnceCell<T> {
         /// Creates a new empty cell.
+        ///
+        /// # Example
+        /// ```
+        /// use once_cell::unsync::OnceCell;
+        ///
+        /// let mut cell: OnceCell<u32> = OnceCell::new();
+        /// cell.set(92).unwrap();
+        /// cell = OnceCell::new();
+        /// ```
         pub const fn new() -> OnceCell<T> {
             OnceCell { inner: UnsafeCell::new(None) }
         }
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -457,15 +466,7 @@ pub mod unsync {
...
         /// This method is allowed to violate the invariant of writing to a `OnceCell`
         /// at most once because it requires `&mut` access to `self`. As with all
-        /// interior mutability, `&mut` access permits arbitrary modification:
-        ///
-        /// ```
-        /// use once_cell::unsync::OnceCell;
-        ///
-        /// let mut cell: OnceCell<u32> = OnceCell::new();
-        /// cell.set(92).unwrap();
-        /// cell = OnceCell::new();
```
matklad, line 467: I think the example is where it needs to be, but is incomplete. Let's change it to 

```rust
use once_cell::unsync::OnceCell;

let mut cell: OnceCell<u32> = OnceCell::new();
cell.set(92).unwrap();
*cell.get_mut().unwrap() = 93;
assert_eq!(cell.get(), Some(&93));

```

matklad:
bors r+

Thanks!

src/lib.rs
muramasa8191: Sounds good. I fixed like that in [a361.](https://github.com/matklad/once_cell/pull/180/commits/a36132d4513f7316576a3b45ddb768210f0825ab)

PR #183, opened 2022-06-21
Add a `Lazy::get` function, similar to `OnceCell::get`

glandium:


matklad:
Makes sense to have this! Could you also add an entry to the changelog and bump the version in Cargo.toml, so that this gets released to crates.io by CI? 

bors d+

PR #184, opened 2022-06-26
Miri has gained some more features

RalfJung:
Barriers, concurrency, and spinloops should all work now.


matklad:
That's just so awesome!

bors r+

PR #185, opened 2022-06-28
Keep provenance intact by avoiding ptr-int-ptr

saethlin:
`once_cell` is an extremely widely-used crate, so it would be very nice if it conformed to the stricted/simplest/checkable model we have for aliasing in Rust. To do this, we need to avoid creating a pointer from an integer by cast or transmute. Pointers created this way can be valid, but are a nightmare for a checker like Miri. The situation is generally explained by these docs: https://doc.rust-lang.org/nightly/std/ptr/fn.from_exposed_addr.html

This implementation is mostly gross because all the APIs that would make this ergonomic are behind `#![feature(strict_provenance)]`. (in particular, it would be great to have `map_addr` here)

### DIFF
--- a/src/imp_std.rs
+++ b/src/imp_std.rs
@@ -3,6 +3,9 @@
 //   * no poisoning
 //   * init function can fail
 
+// Our polyfills collide with the strict_provenance feature
+#![allow(unstable_name_collisions)]
+
 use std::{
     cell::{Cell, UnsafeCell},
     hint::unreachable_unchecked,
@@ -145,8 +148,8 @@ impl<T> OnceCell<T> {
 const INCOMPLETE: usize = 0x0;
 const RUNNING: usize = 0x1;
 const COMPLETE: usize = 0x2;
-const INCOMPLETE_PTR: *mut Waiter = 0x0 as *mut Waiter;
-const COMPLETE_PTR: *mut Waiter = 0x2 as *mut Waiter;
+const INCOMPLETE_PTR: *mut Waiter = INCOMPLETE as *mut Waiter;
+const COMPLETE_PTR: *mut Waiter = COMPLETE as *mut Waiter;
 
 // Mask to learn about the state. All other bits are the queue of waiters if
 // this is in the RUNNING state.
@@ -171,11 +174,11 @@ impl Drop for Guard<'_> {
     fn drop(&mut self) {
         let queue = self.queue.swap(self.new_queue, Ordering::AcqRel);
 
-        let state = queue as usize & STATE_MASK;
-        assert_eq!(state, RUNNING as usize);
+        let state = queue.addr() & STATE_MASK;
+        assert_eq!(state, RUNNING);
 
         unsafe {
-            let mut waiter = (queue as *mut u8).wrapping_sub(state) as *mut Waiter;
+            let mut waiter = queue.map_addr(|q| q & !STATE_MASK);
             while !waiter.is_null() {
                 let next = (*waiter).next;
                 let thread = (*waiter).thread.take().unwrap();
@@ -198,14 +201,13 @@ fn initialize_or_wait(queue: &AtomicPtr<Waiter>, mut init: Option<&mut dyn FnMut
     let mut curr_queue = queue.load(Ordering::Acquire);
 
     loop {
-        let curr_state = curr_queue as usize & STATE_MASK;
+        let curr_state = curr_queue.addr() & STATE_MASK;
         match (curr_state, &mut init) {
             (COMPLETE, _) => return,
             (INCOMPLETE, Some(init)) => {
                 let exchange = queue.compare_exchange(
                     curr_queue,
-                    (curr_queue as *mut u8).wrapping_sub(curr_state).wrapping_add(RUNNING)
-                        as *mut Waiter,
+                    curr_queue.map_addr(|q| (q & !STATE_MASK) | RUNNING),
                     Ordering::Acquire,
                     Ordering::Acquire,
                 );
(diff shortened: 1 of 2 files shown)


src/imp_std.rs
```diff
@@ -145,6 +145,9 @@ impl<T> OnceCell<T> {
 const INCOMPLETE: usize = 0x0;
 const RUNNING: usize = 0x1;
 const COMPLETE: usize = 0x2;
+const INCOMPLETE_PTR: *mut Waiter = core::ptr::null_mut();
+//const RUNNING_PTR: *mut Waiter = core::ptr::null_mut::<Waiter>().wrapping_add(0x1);
+const COMPLETE_PTR: *mut Waiter = core::ptr::null_mut::<u8>().wrapping_add(0x2) as *mut Waiter;
```
RalfJung, line 150: This should work in Rust 1.36
```rust
const COMPLETE_PTR: *mut Waiter = {
    union Transmute { i: usize, p: *mut Waiter }
    unsafe { Transmute { i: 0x2 }.p }
};
```

saethlin:
When running the tests locally with `MIRIFLAGS=-Zmiri-disable-isolation cargo miri test --all-features` Miri ends up in a deadlock. It looks like this doesn't reproduce in CI. 🤷

matklad:
>. (in particular, it would be great to have map_addr here)

Could you add a `map_addr` as a free-standing function ("polyfill")?

src/imp_std.rs
```diff
@@ -145,6 +145,8 @@ impl<T> OnceCell<T> {
 const INCOMPLETE: usize = 0x0;
 const RUNNING: usize = 0x1;
 const COMPLETE: usize = 0x2;
+const INCOMPLETE_PTR: *mut Waiter = 0x0 as *mut Waiter;
```
RalfJung, line 148:
```suggestion
const INCOMPLETE_PTR: *mut Waiter = INCOMPLETE as *mut Waiter;
```

src/imp_std.rs
```diff
@@ -145,6 +145,8 @@ impl<T> OnceCell<T> {
...
+const COMPLETE_PTR: *mut Waiter = 0x2 as *mut Waiter;
```
RalfJung, line 149:
```suggestion
const COMPLETE_PTR: *mut Waiter = COMPLETE as *mut Waiter;
```

src/imp_std.rs
```diff
@@ -224,24 +227,24 @@ fn initialize_or_wait(queue: &AtomicUsize, mut init: Option<&mut dyn FnMut() ->
...
         let exchange = queue.compare_exchange(
             curr_queue,
-            me | curr_state,
+            me.map_addr(|q| q | curr_state),
             Ordering::Release,
             Ordering::Relaxed,
         );
         if let Err(new_queue) = exchange {
-            if new_queue & STATE_MASK != curr_state {
+            if new_queue as usize & STATE_MASK != curr_state {
```
RalfJung, line 247:
```suggestion
            if new_queue.addr() & STATE_MASK != curr_state {
```

src/imp_std.rs
```diff
@@ -224,24 +227,24 @@ fn initialize_or_wait(queue: &AtomicUsize, mut init: Option<&mut dyn FnMut() ->
     }
 }
 
-fn wait(queue: &AtomicUsize, mut curr_queue: usize) {
-    let curr_state = curr_queue & STATE_MASK;
+fn wait(queue: &AtomicPtr<Waiter>, mut curr_queue: *mut Waiter) {
+    let curr_state = curr_queue as usize & STATE_MASK;
```
RalfJung, line 231:
```suggestion
    let curr_state = curr_queue.addr() & STATE_MASK;
```

src/imp_std.rs
```diff
@@ -191,27 +194,27 @@ impl Drop for Guard<'_> {
 //
 // Note: this is intentionally monomorphic
 #[inline(never)]
-fn initialize_or_wait(queue: &AtomicUsize, mut init: Option<&mut dyn FnMut() -> bool>) {
+fn initialize_or_wait(queue: &AtomicPtr<Waiter>, mut init: Option<&mut dyn FnMut() -> bool>) {
     let mut curr_queue = queue.load(Ordering::Acquire);
 
     loop {
-        let curr_state = curr_queue & STATE_MASK;
+        let curr_state = curr_queue as usize & STATE_MASK;
```
RalfJung, line 201:
```suggestion
        let curr_state = curr_queue.addr() & STATE_MASK;
```

src/imp_std.rs
```diff
@@ -156,23 +158,24 @@ const STATE_MASK: usize = 0x3;
...
+    new_queue: *mut Waiter,
 }
 
 impl Drop for Guard<'_> {
     fn drop(&mut self) {
         let queue = self.queue.swap(self.new_queue, Ordering::AcqRel);
 
-        assert_eq!(queue & STATE_MASK, RUNNING);
+        let state = queue as usize & STATE_MASK;
+        assert_eq!(state, RUNNING as usize);
```
RalfJung, line 175:
```suggestion
        let state = queue.addr() & STATE_MASK;
        assert_eq!(state, RUNNING);
```

RalfJung (commented):
if we have that polyfill we might as well use it? :D

LoganDark:
Is there a reason to be pasting in sptr rather than depending on it?

Nilstrieb:
I assume it's to avoid having an extra dependency, as `sptr` contains more code than is required here, which would have a negative impact on build times.

LoganDark:
> I assume it's to avoid having an extra dependency, as `sptr` contains more code than is required here, which would have a negative impact on build times.

`sptr` is a relatively small crate compared to what you'd usually blame for inflating compile times; a proper benchmark is required to reach a proper conclusion. I'd assume the bulk of the overhead would be the dependency itself (i.e. cargo having to download and compile it separately), not the code inside, but `cargo` is already very good at compiling large dependency trees.

matklad:
bors r+

I don't think it's worth it to go from 0 deps to >0 deps just for this

LoganDark:
Well, at least this is getting merged.

matklad:
Released as https://crates.io/crates/once_cell/1.13.1

PR #188, opened 2022-07-20
update `regex!` to support multiple regexes and `expr`s

vidhanio:
adding multiple regexes in the macro will return a tuple with all the regexes, and also allow using an `expr` to compile into the regex. this leaves the original api untouched, only adding more features.


matklad:
This is an example, so I’d rather keep it simple rather than featerful.

vidhanio:
i guess that does make sense, but in my opinion it doesn't add too much visual complexity to the documentation (although i admit the description could maybe use some work).

PR #190, opened 2022-08-13
correct minor typo and bumped atomic-polyfill

jqnatividad:


matklad:
released as https://crates.io/crates/once_cell/1.13.1

taiki-e:
FYI, this was a breaking change for users who use the atomic-polyfill feature. (atomic-polyfill itself is not used in the once_cell public API, but the mechanism to implement a critical session implementation in the atomic-polyfill depends on versions of the atomic-polyfill and the critical-section (used in atomic-polyfill), so this can cause a compile error. I can provide reproduction if needed.)

That said, I think this breaking change is justified because atomic-polyfill 1.0 has mitigated/fixed several soundness issues.

Issue #113, opened 2020-08-20
Lazy force_mut?


camsteffen:
I see that `DerefMut` was added to `Lazy`. Should there also be `force_mut` then?

matklad:
Yeah, we probably should add this here and to the std

danielSanchezQ:
What is the state of this? I just realised I would need a `get_mut_or_init`. To get a `&mut` after initialisation.

matklad:
waiting for someone to submit a PR with implementaiton, docs & tests :)

danielSanchezQ:
> waiting for someone to submit a PR with implementaiton, docs & tests :)

Count me in then! Would need some guidance probably. But expect a PR soon 😄

matklad:
This exists now

PR #193, opened 2022-09-01
`Lazy` mutable  API extension

danielSanchezQ:
As requested in #113.  Also added explicit `get_mut`. 

Implement:

- [x] `async::Lazy::force_mut`
- [x] `async::Lazy::get_mut`
- [x] `sync::Lazy::force_mut`
- [x] `sync::Lazy::get_mut`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@
 - Add extension to `async` and `sync` `OnceCell` mut API:
   - `get_mut_or_init`
   - `get_mut_or_try_init`
+- Add extension to `async` and `sync` `Lazy` mut API:
+  - `force_mut`
+  - `get_mut`
 
 
 ## 1.13.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ## 1.14.0
 
-- Add extension to `async` and `sync` `Lazy` mut API:
+- Add extension to `unsync` and `sync` `Lazy` mut API:
   - `force_mut`
   - `get_mut`
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,6 @@
 
 ## 1.14.0
 
-- Add extension to `async` and `sync` `OnceCell` mut API:
-  - `get_mut_or_init`
-  - `get_mut_or_try_init`
 - Add extension to `async` and `sync` `Lazy` mut API:
   - `force_mut`
   - `get_mut`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@
 
 -
 
+## 1.14.0
+
+- Add extension to `async` and `sync` `OnceCell` mut API:
+  - `get_mut_or_init`
+  - `get_mut_or_try_init`
+
+
 ## 1.13.1
 
 - Make implementation compliant with [strict provenance](https://github.com/rust-lang/rust/issues/95228).
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "once_cell"
-version = "1.13.1"
+version = "1.14.0"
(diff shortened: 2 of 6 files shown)


danielSanchezQ:
It do not really makes sense to add this API to `sync` righ @matklad ?
I also don't know if there is need of more documentation/examples for this. If so, point me to it please 😃

matklad:
>It do not really makes sense to add this API to sync righ @matklad ?

oups, missed this Q, sorry. 

No, I think it *does* make sense to have this for sync as well, the same as `sync::OnceCell::get_mut`

danielSanchezQ:
> > It do not really makes sense to add this API to sync righ @matklad ?
> 
> oups, missed this Q, sorry.
> 
> No, I think it _does_ make sense to have this for sync as well, the same as `sync::OnceCell::get_mut`

Cool, will take a look at how `sync::OnceCell::get_mut` is implemented and add it.

danielSanchezQ:
@matklad I added same mut API for `OnceCell`, which is then used for `sync::Lazy`. I am having an issue enabling `get_mut_unchecked` it may have to deal with some feature enabling but I can not get it right. Would you please take a quick look?

matklad:
I feel that in the latest iteration it maybe scope-crept to far. I don't think we need to extend `OnceCell` APIs -- just `get_mut().unwrap()` should be enough. If the call-site really cares about not having a panic branch there, it can use `.unwrap_unchecked`.

danielSanchezQ:
> I feel that in the latest iteration it maybe scope-crept to far. I don't think we need to extend `OnceCell` APIs -- just `get_mut().unwrap()` should be enough. If the call-site really cares about not having a panic branch there, it can use `.unwrap_unchecked`.

@matkad, you mean all the implementations or just the `get_mut_unchecked`? 
I agree that `get_mut_unchecked` may be too much, but I added it as I wanted to replicate the API. But totally down to remote it again 😄

matklad:
I think we only need to add these four APIs

* unsync::Lazy::force_mut
* unsync::Lazy::get_mut
* sync::Lazy::force_mut
* sync::Lazy::get_mut

And `OnceCell::get_mut` should be enough to implement those.

danielSanchezQ:
> I think we only need to add these four APIs
> 
> * unsync::Lazy::force_mut
> * unsync::Lazy::get_mut
> * sync::Lazy::force_mut
> * sync::Lazy::get_mut
> 
> And `OnceCell::get_mut` should be enough to implement those.

Not to discuss here probably. But having the option to get the `&mut` as well as the `&` when getting or initializing end up solving the issue in where you end up with something like:

```rust
let cell = OnceCell::new();
cell.get_or_init(|| String::new("foo"));
let mut ref = cell.get_mut().unwrap();
```

and instead keep the api as expected:


```rust
let cell = OnceCell::new();
let mut ref = cell.get_mut_or_init(|| String::new("foo"));
```

It seems not too much, but bump into that already (and today once again 😅 ).

danielSanchezQ:
@matklad , I left just the `Lazy` items as discussed. I will open an issue+PR with the `OnceCell` ones so we do not mix discussions 😄

CHANGELOG.md
```diff
@@ -4,6 +4,13 @@
 
 -
 
+## 1.14.0
+
+- Add extension to `async` and `sync` `Lazy` mut API:
```
matklad, line 9:
```suggestion
- Add extension to `unsync` and `sync` `Lazy` mut API:
```

matklad:
bors r+

Thanks!

matkad:
Hi,

don’t really know how my email got onto this distribution list :-) being a programmer myself, I keep fingers crossed for the success of your project.
Let me know when it’s done, but please I don’t need to see all the small updates :D 

Cheers,
Mateusz

matklad:
matkad sorry, I think that was a typo in a mention due to our github logins having a hamming distance of one: https://github.com/matkad vs https://github.com/matklad

Issue #61, opened 2019-10-08
spin-based implementation for no_std


matklad:
We currently don't provide `sync` module in no_std, as it requires OS-level support for blocking. What lazy_static does in this case is that it has a `spin` feature, which replaces locking with spinning. 

I think just silently replacing OS mutexes with busy waiting is the wrong approach, this is an important distinction which should be reflected in the type system. 

So, we should instead add an opt-in `spin` module, which has the same API as `sync`, but is based on the `spin` crate. That is, with both `std` and `spin` features enabled, the user should be able to use both `once_cell::sync` and `once_cell::spin`.

richardanaya:
I just ran into needing this tonight and felt bad having to turn to lazy static :( thanks for opening this! once_cell FTW!

josephlr:
So notes on why we might want this (both in `once_cell` and eventually merged into `libstd`/`libcore`):
  - `ring` needs a `no_std` compatible replacement for `spin-rs` (which is unmaintained), https://github.com/briansmith/ring/issues/921 
  - `getrandom` (a dependency of `rand`) needs one-time initialization for holding onto file handles, checking CPUID, checking OS support for various functionality, etc... Right now, we [implement this on our own](https://github.com/rust-random/getrandom/blob/master/src/util.rs#L11-L83) but ideally we wouldn't have this custom implementation. As we want to eventually make `getrandom` part of the `libstd` (see https://github.com/rust-lang/rust/issues/62079 and https://github.com/rust-lang/rust/pull/62082), it cannot depend on any `std` features. So using spinlocks is the best bet.

matklad:
>So using spinlocks is the best bet.

I think I disagree with this. Using spin-locks if there's a real operating system around seems bad. If two threads race to run initialization, one thread enters a critical section and is scheduled out of the CPU, the other will be busy waiting for a long time. Moreover, if the first thread is a low-priority one, and the second one has a high priority, we get priority inversion! 

Moreover, I don't think std uses blocking at all at the moment when getting random data? I think std only needs randomness for hash maps (is this true?) and there, it uses tls for caching:

https://github.com/rust-lang/rust/blob/e380efa5ecdef714dad72c473fc0933ff4d59283/src/libstd/collections/hash/map.rs#L2459-L2461

Could the `getrandom` be designed in such a way that it's the client who manages the state?

Something like this

```rust
pub struct SysRandom { ... }

impl SysRandom {
    pub fn init() -> SysRandom { ... }
[...]

PR #195, opened 2022-09-06
Add `no_std` implementation based on `critical-section`.

reitermarkus:
Add implementation based on `critical-section` for embedded targets.

This started out using `critical_section::Mutex`, but tests were failing since `critical_section::with` blocks all threads during initialization, so I changed it to be basically the same as `imp_pl` without the `parking_lot` parts, since `atomic-polyfill` is based on `critical-section`.

Depends on https://github.com/rust-embedded/critical-section/pull/26.

### DIFF
--- a/src/imp_cs.rs
+++ b/src/imp_cs.rs
@@ -47,8 +47,6 @@ impl<T> OnceCell<T> {
             let cell = self.value.borrow(cs);
             cell.get_or_try_init(f).map(|_| {
                 self.initialized.store(true, Ordering::Release);
-
-                ()
             })
         })
     }
--- a/src/imp_cs.rs
+++ b/src/imp_cs.rs
@@ -1,22 +1,12 @@
-#[cfg(feature = "atomic-polyfill")]
-use atomic_polyfill as atomic;
-#[cfg(not(feature = "atomic-polyfill"))]
-use core::sync::atomic;
-
-use atomic::{AtomicU8, Ordering};
-
 use core::panic::{RefUnwindSafe, UnwindSafe};
+
+use critical_section::{CriticalSection, Mutex};
 
 use crate::unsync::OnceCell as UnsyncOnceCell;
 
 pub(crate) struct OnceCell<T> {
-    state: AtomicU8,
-    value: UnsyncOnceCell<T>,
+    value: Mutex<UnsyncOnceCell<T>>,
 }
-
-const INCOMPLETE: u8 = 0;
-const RUNNING: u8 = 1;
-const COMPLETE: u8 = 2;
 
 // Why do we need `T: Send`?
 // Thread A creates a `OnceCell` and shares it with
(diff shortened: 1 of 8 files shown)


src/imp_cs.rs
```diff
@@ -0,0 +1,131 @@
...
+        {
+            Ok(_) => {
+                let mut guard = Guard { state, new_state: INCOMPLETE };
+                if init() {
+                    guard.new_state = COMPLETE;
+                }
+                return;
+            }
+            Err(COMPLETE) => return,
+            Err(RUNNING) | Err(INCOMPLETE) => core::hint::spin_loop(),
```
matklad, line 124: I don't think we should have a spin-lock anywhere, especially given that we are using critical section anyway, which gives us locking.

matklad (commented):
:thinking: so, historically I've been pretty opposed to "just works" implementation on no_std, as how exactly one does locking is pretty important, and spinlocks is decidedly not a way to go, there's some discussion in https://github.com/matklad/once_cell/issues/61. 

The approach roughly like this could work though:

* if we use a critical section rather than spinning, that counts as proper blocking in at least some contexts
* the experience is that the user has to explicitly enable the `critical_section` feature, so they are opting into this behavior

So, yeah, I think I am generally on board with it. However, it is important that we don't actually spin here (the current implementation does spin).

reitermarkus:
Okay, so I have a different implementation which doesn't spin and simply completely blocks all threads during `initialize`, but this causes `cargo test` to deadlock unless using `--test-threads 1`.

reitermarkus:
@matklad, can you have another look?

src/imp_cs.rs
```diff
@@ -0,0 +1,78 @@
+use core::panic::{RefUnwindSafe, UnwindSafe};
+
+use atomic_polyfill::{AtomicBool, Ordering};
+use critical_section::{CriticalSection, Mutex};
+
+use crate::unsync::OnceCell as UnsyncOnceCell;
+
+pub(crate) struct OnceCell<T> {
+    initialized: AtomicBool,
+    value: Mutex<UnsyncOnceCell<T>>,
```
matklad, line 10: On mobile, so not a proper review, but I am wondering why this needs to be a OnceCell? It seems like an Option should be enough perhaps?

matklad: Ah, the docs explain that nicely:

>It does not provide interior mutability.

worth adding a comment about this!

reitermarkus: Ah, yes, that too. And also, using `unsync::OnceCell` inside allows re-using the `get_or_try_init` logic.

src/lib.rs
```diff
@@ -329,29 +329,38 @@
...
+#[cfg(not(feature = "std"))]
+use core as lib;
+#[cfg(feature = "std")]
+use std as lib;
+
+use lib::{
+    cell::{Cell, UnsafeCell},
+    fmt, mem,
+    ops::{Deref, DerefMut},
+    panic::{RefUnwindSafe, UnwindSafe},
```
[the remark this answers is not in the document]
reitermarkus: I think that's why I was using that `use std as lib` workaround. So `MSRV` for the `std` feature stays the same.

matklad:
lgtm!

bors r+

PR #198, opened 2022-09-15
Release 1.15.0

matklad:


matklad:
NB: this bump MSRV to 1.56.0. I think this should be fine, but, if you have specific objection to that, I'd love to hear. 

This PR pushes 1.15.0-pre.1, I'll upgrade that to 1.15.0 proper next week.

matklad:
re-reading https://github.com/rust-lang/libs-team/issues/72, I think my stance on that is that once-cell intends to support projects which generally keep with up-to-date Rust, while also providing generous comparability window. It is a non-goal to support use-cases which have to use old compilers without intention to upgrade. 

For debian-stable-like specifically, I think the reasonable choices are: 

* install a more recent Rust using rustup
* if using old compiler is required, use old version of libraries as well
* if combination of old compiler and new library is required, it's on the consumer to patch the library on their side

This is based primarily on two observations: 

* overall Rust ecosystem policy is closer to "live at head". When Rust policy clashes with distro policy, Rust policy is preferred
* practically, it seems relatively few people are stuck with as old Rust as debian stable, so this shouldn't have a big practical impact. And, arguably, those people who *are* impacted could be advised to find a way to use a supported version of Rust

matklad:
From tokio discussion: 

Tokio has MSRV of 6 month, but it *also* has a 12 month LTS, which means that effective MSRV for tokio is 18. 

I don't think it makes sense for once_cell to chase downstream LTS promises (which is a restatement of the policy above that says that once_cell generally caters to users who aspire to use a somewhat up-to-date compiler). 

So, I'll cut a release!

PR #212, opened 2022-12-23
Fixed `LateInit` recipe stack overflow

jack-cooper:
The given `LateInit` recipe works up until the print statement, which causes a stack overflow as the derived `Debug` implementation tries to write out:
```rs
LateInit { cell: OnceCell(A { b: LateInit { cell: OnceCell(B { a: LateInit // repeats endlessly from here
```
To fix this, I've added in a rudimentary `Debug` impl.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -206,11 +206,8 @@
 //!
 //!
 //! ```
-//! use std::fmt;
-//!
 //! use once_cell::sync::OnceCell;
 //!
-//! #[derive(Debug)]
 //! pub struct LateInit<T> { cell: OnceCell<T> }
 //!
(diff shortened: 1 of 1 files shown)


matklad:
:thinking: yeah, ideally even silly examples would do something reasonable. I'd like to aviod adding custom code for debug here -- destracts from the thrust of the example. Can we rework this somehow to keep it simple? maybee use something like `let _a = a.b.a.b.a` instead of print?

jack-cooper:
Sounds sensible to me! Most recent commit removes the `Debug` impls.

matklad:
Nice, thanks!

bors r+

PR #213, opened 2022-12-25
Add once_cell::race::OnceRef

gh2o:
Like OnceBox, but stores a shared reference instead of a Box.

Mainly useful for `#![no_std]` environments where we can use it to store a `&'static T`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 -
 
+## 1.17.0
+
+- Add `race::OnceRef` for storing a `&'a T`.
+
 ## 1.16.0
 
 - Add `no_std` implementation based on `critical-section`,
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "once_cell"
-version = "1.16.0"
+version = "1.17.0"
 authors = ["Aleksey Kladov <aleksey.kladov@gmail.com>"]
 license = "MIT OR Apache-2.0"
 edition = "2021"
--- a/src/race.rs
+++ b/src/race.rs
@@ -25,6 +25,8 @@ use atomic_polyfill as atomic;
 use core::sync::atomic;
 
 use atomic::{AtomicUsize, Ordering};
+use core::cell::UnsafeCell;
+use core::marker::PhantomData;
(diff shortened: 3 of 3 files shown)


matklad:
Makes sense to have this, yeah! Could you express this in onterms of `AtomicUsize` though in the implementation?

```rust
pub struct OnceRef<'a, T> {
    inner: OnceNonZeroUsize,
    ghost: PhantomData<Option<&'a T>>,
}
```

We don't do that for OnceBox because it needs some custom drop code, but for OnceRef I think it should work!

I am also a tiny bit worried about phantom data and variance here. If you can a compile_fail test that we can't use this to somehow shorten the lifetime in an inapropriate way, that would be nice!

:thinking: actually, it does seem that we also need to do something about the lifetimes here, the current code seems to be unsound:


```rust
#[test]
fn lifetimes_test() {
    let x = 1;
    let l = OnceRef::new();
    if false {
        l.set(&x).unwrap();
    }
    {
        let y = 2;
        let r = OnceRef::new();
        r.set(&y).unwrap();
        swap(&l, &r);
    }
    eprintln!("uaf: {}", *l.get().unwrap());
    fn swap<'a>(l: &OnceRef<'a, i32>, r: &OnceRef<'a, i32>) {
        let lr = l.get();
        let rr = r.get();
        if let Some(rr) = rr {
            let _ = l.set(rr);
        }
        if let Some(lr) = lr {
            let _ = l.set(lr);
        }
    }
}
```

gh2o:
Good catch on the lifetime issue. I've updated the code to use `OnceNonZeroUsize`, and changed `ghost` to `PhantomData<UnsafeCell<&'a T>>` so that your test now fails with `error[E0597]: `y` does not live long enough`.

matklad:
Perfect, thanks!

could you also add an entry to change log, and bump a version in Cargo.toml, so that ci infra does a release on merge?

gh2o:
> could you also add an entry to change log, and bump a version in Cargo.toml, so that ci infra does a release on merge?

Done. Also updated `set` to return `()` on error similar to the other `Once*` types other than `OnceBox` (since `&'a T` is `Copy`), and made `get_or_init` and `get_or_try_init` return `&'a T` instead of `&'self T`.

PR #214, opened 2023-01-02
Explain safety of `unsync::OnceCell::get(&self)` in more detail

Enselic:
Background: I'm working on improving my unsafe Rust skills, and figuring out how `once_cell` works is one way to do that. By more elaborately explaining why `unsync::OnceCell::get(&self)` is safely implemented, I hope to help others increase their understanding of unsafe Rust as well.

(It was not obvious to me at first why `unsync::OnceCell::get(&self)` was safe. But I think I figured it out now.)


matklad:
bors r+

Thanks!

PR #219, opened 2023-02-14
Use AtomicPtr for race::OnceRef to avoid ptr-int-ptr

Imberflur:
Same rationale as https://github.com/matklad/once_cell/pull/185

Potentially, a `OncePtr` abstraction could be factored out and used in both `OnceBox` and `OnceRef`.


matklad:
Yeah, this makes sense!

bors r+

PR #227, opened 2023-04-12
Add from_fn

dspyz-matician:
Sometimes the compiler fails to infer the type of `T` in a `Lazy`. This helps the compiler figure it out from the closure.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -729,6 +729,25 @@ pub mod unsync {
     }
 
     impl<T, F: FnOnce() -> T> Lazy<T, F> {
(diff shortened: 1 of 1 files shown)


matklad:
Hm, I am torn a bit... Ideally, we'd just add this bound to new, like we do for std. 

That would be an incompatible change though (the reason why we don't have this bound is that it was impossible until some relatively recent version of Rust)

I don't want to break semver by adding the bound retroactively, but I am also not sure that adding a whole new method here is justified...

matklad:
To not let this hang indefinitely, I decided to close this: to me it seems that smaller API surface outweighs reduced annotation burden.

Issue #167, opened 2021-12-06
Feature request: Make `Lazy<T, F>` covariant in `F`


jorendorff:
I have some code where we'd like to use `Lazy` values as fields of structs, which means boxing closures:

```rust
type AnyLazy<'a, T> = `Lazy<T, Box<dyn FnOnce() -> T + 'a>>;
```

The problem is that this type alias is invariant in `'a`, because `Lazy<T, F>` is invariant in `F`. So actually using this type causes insurmountable lifetime issues in our code.

I hesitate to ask for this, because the only way to get it is to lie a lot to Rust about what's inside the `Cell`. The API would be safe, but I don't see a way to do it without a fairly gross amount of unsafe code.

jorendorff:
If this is something you'd consider, let me know. I can give it a shot.

matklad:
I think we should have it. I don't particularly mind unsafe, as long as it is sound. One problem here is that it'll be hard to check soundness for me. I guess, let's do the following: 

* you write PR with comments and tests
* I get up-to-speed with respect to variance and double check that the logic is sound
* at the end, we rope in RalfJung / r/rust to poke holes in our reasoning?

matklad:
We also should fix https://doc.rust-lang.org/std/lazy/struct.Lazy.html in the same way -- for std types, it's very important that they are pedantically correct in terms of variance and such.

danielhenrymantilla:
For what is worth, since the api w.r.t. `F` is _ownership_-based exclusively (no `&` or `&mut` on the (value of type) `F`), it means covariance is necessarily fine 🙂

douglas-raillard-arm:
Drive by comment: I've hit a similar issue, for which I could not find a fully satisfactory answer because it seems impossible to get interior mutability and covariance at the same time for owned data.

> We also should fix https://doc.rust-lang.org/std/lazy/struct.Lazy.html in the same way -- for std types, it's very important that they are pedantically correct in terms of variance and such.

I fully agree. Even `UnsafeCell<T>` is invariant in `T`, which makes it basically impossible to use as the basis of a memoization system even if it's perfectly fine (i.e. replacing the original value by either a value of the exact same lifetime or by `'static` ). This is a bummer especially since I'd expect `UnsafeCell` to give me maximum (unsafe) control, given it's the only primitive that can be used.

endorpersand:
[...]

PR #230, opened 2023-05-27
Proof of Concept: making `Lazy<T, F>` covariant in `F`

danielhenrymantilla:
Proof of Concept of my idea from https://github.com/matklad/once_cell/issues/167#issuecomment-1564727077

  - `CovariantCellOption<F>` is soundly covariant since the `F` value is never overwritten.

If this PoC were to be merged, it would fix https://github.com/matklad/once_cell/issues/167.

Although it does so at the cost of loss of discriminant elision of `Option<F>`, which means it makes `Lazy`s bigger. This could be deemed too high a cost for a somewhat niche feature.

  - an alternative could be to offer a `LazyCovariantCtor<T, F>` type which would thereby let users pick their preferred flavor.

### DIFF
--- a/src/covariant_cell_option.rs
+++ b/src/covariant_cell_option.rs
@@ -1,0 +1,59 @@
+//! A `Cell<Option<F>>`, but covariant, at the cost of a very reduced API.
+//!
+//! (the idea being: it starts as `Some(F)`, and the only `&`-based operation
+//! is `take()`ing it. This guarantees covariance to be fine, since the `F`
+//! value is never overwritten).
+
+use ::core::{cell::Cell, mem::ManuallyDrop};
+
+pub struct CovariantCellOption<F> {
+    /// Invariant: if this is `true` then `value` must contain a non-dropped `F`;
+    is_some: Cell<bool>,
+    value: ManuallyDrop<F>,
+}
+
+impl<F> CovariantCellOption<F> {
+    pub const fn some(value: F) -> Self {
+        Self { is_some: Cell::new(true), value: ManuallyDrop::new(value) }
+    }
+
+    pub fn into_inner(self) -> Option<F> {
+        // Small optimization: disable drop glue so as not to have to overwrite `is_some`.
+        let mut this = ManuallyDrop::new(self);
+        let is_some = this.is_some.get_mut();
+        is_some.then(|| unsafe {
+            // SAFETY: as per the invariant, we can use `value`. We can also *consume* it by doing:
+            // *is_some = false;
+            // but we actually don't even need to do it since we don't use `this` anymore.
+            ManuallyDrop::take(&mut this.value)
+        })
+    }
+
+    pub fn take(&self) -> Option<F> {
+        self.is_some.get().then(|| unsafe {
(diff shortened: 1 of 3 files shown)


matklad:
TBH, I am off two minds here, as the following concerns conflict:

* as a library author, I want to pull as much complexity as posible into the lib's code, so that, from the user's perspective, things "just work".
* at the same time, I very much don't want to "go against the grain of the language", and implement clever work-arounds.

I guess, on balance, yeah, we should do that. 

But before, I am curious, can we optimize this a bit? In particular, the current impl,

```rust
    pub struct Lazy<T, F = fn() -> T> {
        cell: OnceCell<T>,
        init: Cell<Option<F>>,
    }
```

isn't really what it _should_ be, as we have two flags here one in `OnceCell`, and one in `Option`. And we don't share the space between `T` and `F`. 

So I think ideally we want something like

```rust
struct Lazy<T, F> {
    state: enum { Init, Uninit, Poison },
    data: union {
        f: F,
        t: T,
        poison: (),
    }
}
```

can we make that work pedantically correct wrt variance?

matklad:
I guess, as another consideration here, give that `std::sync::LazyLock` is almost stable, I'll probably change what this crate is about, and move it from "stable foundation" to "fancy extras", as the foundational role would be subsumed by std. From this perspective, doing something more creative here also makes sense.

danielhenrymantilla:
> can we make that work pedantically correct wrt variance?

Yes, by slightly amending it 🙂:

```rs
use ManuallyDrop as MD; // <- necessary red tape

struct Lazy<T, F> {
    state: SomeMutWrapper<enum { Init, Uninit, Poison }>, // Uninit=can read `f`, Init=can read 't', Poison=cannot read either
    data: union {
        f: MD<F>,
        t: UnsafeCell<MD<T>>, // 👈 `Unsafe` Cell only here
    }
}
```

  - _e.g._, the `impl Drop`:

    ```rs
    match self.state.get_mut() {
        Init => unsafe { MD::drop(&mut *self.data.t.get_mut()) },
        Uninit => unsafe { MD::drop(&mut self.data.f) },
        Poison => {},
    }
    ```

  - `take`ing the `f` would then be a matter of setting the `state` to `Poison`, then `<*const _>::read()`ing the `.f` field like this very PR does, and writing that value to `.t`, and then setting the `state` to `Init`.

This has the right variance w.r.t. both `F`  and `T` and would indeed allow squashing the `is_some` flag of `.f` in my PR within `state`. The only drawback is that it requires reïmplementing the flag-handling for `Lazy`, rather than being able to rely on the `OnceCell<T>` already existing abstraction.

  - The advantage, however, for people using `Lazy` with zero-sized `impl Fn` types (_e.g._, within function bodies or using TAITs), it that it would mean `Lazy` would be getting smaller for them as well!

    In other words, this last implementation you've suggested would be both pedantically correct w.r.t. variance, and efficient!

danielhenrymantilla:
I'm thus closing this draft PR, but may be submitting a PR implementing this for `unsync::Lazy` (the "easy" to write implementation of `OnceCell`). From there, I'll let sync-savy people follow-up on this and handle the `sync` flavor thereof

PR #231, opened 2023-05-28
Avoid unnecessary synchronization in `{force,deref}_mut`

### LINKED ISSUE #226: Nit: unnecessary synchronization in `Lazy::force_mut()`
`Self::force()` has to synchronize the access to the raw value and/or its needed initialization, given its `&`-based API, but `force_mut()` does not, thanks to the guarantees of `&mut` access (_e.g._, https://doc.rust-lang.org/stable/std/cell/struct.UnsafeCell.html#method.get_mut).

But the current implementation of `force_mut()` just defers to `force()`:

https://github.com/matklad/once_cell/blob/35148638c54c6233545c65d1a5e09d5ba0661806/src/lib.rs#L767-L770

danielhenrymantilla:
- `DerefMut` was not delegating to `force_mut()` (contary to the by-ref APIs);
  - `Lazy::force_mut` was not taking advantage of exclusive-mutability access, and instead paying shared-mutability access. Mainly, in the `sync` case, it involved atomic operations which are now skipped.

Fixes #226

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -796,8 +796,14 @@ pub mod unsync {
         /// assert_eq!(*lazy, 92);
         /// ```
         pub fn force_mut(this: &mut Lazy<T, F>) -> &mut T {
-            Self::force(this);
-            Self::get_mut(this).unwrap_or_else(|| unreachable!())
+            if this.cell.get_mut().is_none() {
+                let value = match this.init.get_mut().take() {
+                    Some(f) => f(),
+                    None => panic!("Lazy instance has previously been poisoned"),
+                };
+                this.cell = OnceCell::with_value(value);
+            }
+            this.cell.get_mut().unwrap_or_else(|| unreachable!())
         }
 
         /// Gets the reference to the result of this lazy value if
@@ -844,8 +850,7 @@ pub mod unsync {
 
     impl<T, F: FnOnce() -> T> DerefMut for Lazy<T, F> {
         fn deref_mut(&mut self) -> &mut T {
-            Lazy::force(self);
-            self.cell.get_mut().unwrap_or_else(|| unreachable!())
+            Lazy::force_mut(self)
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -1324,8 +1329,14 @@ pub mod sync {
...
-            Self::force(this);
-            Self::get_mut(this).unwrap_or_else(|| unreachable!())
+            if this.cell.get_mut().is_none() {
+                let value = match this.init.get_mut().take() {
+                    Some(f) => f(),
+                    None => panic!("Lazy instance has previously been poisoned"),
+                };
+                this.cell = OnceCell::with_value(value);
+            }
+            this.cell.get_mut().unwrap_or_else(|| unreachable!())
```
danielhenrymantilla, line 1339: (Rewriting this so as to avoiding the double-`get_mut()` on the happy path [runs into `polonius`](https://docs.rs/polonius-the-crab/0.3.1/polonius_the_crab/index.html#rationale-limitations-of-the-nll-borrow-checker))

matklad:
Could you bump a patch version in Cargo.toml and add a short note to the Changelog?

matklad:
bors d+

danielhenrymantilla:
@matklad I won't approve the PR myself since I have also taken the liberty to add a `Cargo.lock.msrv` check (while also updating it), which you may want to review 🙂

xtask/src/main.rs
```diff
@@ -48,7 +48,21 @@ fn main() -> xshell::Result<()> {
...
+                (cp Cargo.lock{{.msrv,}} && cargo +{MSRV} update -w -v && cp Cargo.lock{{.msrv,}})\n\
+                \n\
+                Alternatively, `git apply` the `.patch` below:\
+            "}
+            cmd!(sh, "cargo update -q -w").quiet().run()?;
+            sh.copy_file("Cargo.lock", "Cargo.lock.msrv")?;
+            cmd!(sh, "git --no-pager diff --color=always -- Cargo.lock.msrv").quiet().run()?;
+            return err;
+        }
+        cmd!(sh, "cargo build --locked").run()?;
```
danielhenrymantilla, line 65: Preview:

Screenshot 2023-05-28 at 22 07 17

PR #232, opened 2023-05-29
Added get_mut_unchecked

dynos01:
I added `get_mut_unchecked()` as it might help someone. 

Also, there are two things I'm considering but currently unsure about how to correctly implement:

- Add `wait_mut()`. However mutable reference across threads requires Mutex, but waiting with Mutex acquired will block other threads from initializing the value. Is there a case that this will make sense?
- Add `get_mut_unchecked()` in `imp_cs.rs`. This seems to require modification of the upstream crate `critical_section`, so I skipped it for now.


dynos01:
For point 2, this does compile, but will it work as expected? (I don't have experience with crate `critical-section`, and the definitions of `borrow` and `borrow_mut` [here](https://docs.rs/critical-section/1.1.1/critical_section/struct.Mutex.html#method.borrow_mut) look different)

```
pub(crate) unsafe fn get_mut_unchecked(&mut self) -> &mut T {
    debug_assert!(self.is_initialized());
    // SAFETY: The caller ensures that the value is initialized and access synchronized.
    crate::unwrap_unchecked(self.value.borrow_mut().get_mut().get_mut())
}
```

src/lib.rs
```diff
@@ -1031,6 +1031,19 @@ pub mod sync {
...
+        /// Get the mutable reference to the underlying value, without checking if the
+        /// cell is initialized.
+        ///
+        /// # Safety
+        ///
+        /// Caller must ensure that the cell is in initialized state, and that
+        /// the contents are acquired by (synchronized to) this thread.
+        #[inline]
+        #[cfg(not(feature = "critical_section"))]
+        pub unsafe fn get_mut_unchecked(&self) -> &mut T {
```
matklad, line 1043: What's the concrete use-case for this function? Given `& -> &mut` signature, this seems more or less impossible to use correctly.

dynos01: My initial purpose was to implement `wait_mut`, but I soon realized that `wait_mut` seemed meaningless. But yes, it should be `&mut -> &mut` here. Since there are `get_unchecked` and `get_mut`, will a `get_mut_unchecked` look reasonable?

matklad: I think that’s already available as `get_mut().unwrap_unchecked()`

dynos01: Then what about a shortcut that's self-explanatory? I've added a new commit to fix the function signature, but if you believe this is not necessary, feel free to close this PR. Thanks.

PR #233, opened 2023-06-02
Make `unsync::Lazy<T, F>` covariant in `F`

danielhenrymantilla:
"Continuation" from https://github.com/matklad/once_cell/pull/230, which partially handles https://github.com/matklad/once_cell/issues/167.

  - Incidentally, this also makes `unsync::Lazy` smaller in size in most cases, since `T` and `F` are now sharing the same `union` storage.

The `sync` can basically use the same logic (my PR paves the way for such a follow-up PR), only the whole thing would need to be moved to each of the possible `imp`lementations of `sync::OnceCell`, and special care to synchronization semantics will be in order, which I prefer to let somebody else do.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -716,15 +716,52 @@ pub mod unsync {
     /// //   92
     /// ```
     pub struct Lazy<T, F = fn() -> T> {
-        cell: OnceCell<T>,
-        init: Cell<Option<F>>,
+        state: Cell<State>,
+        data: Data<T, F>,
+    }
+
+    /// Discriminant of the union.
+    #[derive(Debug, Clone, Copy)]
+    enum State {
+        /// `.init` has not been taken yet.
+        Uninit,
+        /// `.init` has been *taken*, but the call has not completed (panic?), so there is no `.value` either.
+        Poisoned,
+        /// `.value` has been set.
+        Value,
+    }
+
+    // One of the rare cases `#[repr(C)]` is not needed for an union ("externally tagged `enum`" pattern)
+    union Data<T, F> {
+        /// Idea: we don't need *mutation* to dispose of `F`, we can just let it be there, inert, behind a MD.
+        ///
+        /// This makes it so we remain covariant in `F`.
+        init: mem::ManuallyDrop<F>,
+        value: mem::ManuallyDrop<UnsafeCell<T>>,
+    }
+
+    impl<T, F> Drop for Lazy<T, F> {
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -716,15 +716,52 @@ pub mod unsync {
     /// //   92
     /// ```
     pub struct Lazy<T, F = fn() -> T> {
-        cell: OnceCell<T>,
-        init: Cell<Option<F>>,
+        state: Cell<State>,
+        data: Data<T, F>,
```
danielhenrymantilla, line 720: Note that we could make `Lazy` even more compact by moving `state` at the beginning of each union variant (and making the `union` be `#[repr(C)]`. If `T = [u8; 3]`, and `F = u16`, this would make `Lazy<T, F>` be `4 bytes` rather than `6`, for instance.

matklad:
bors r+

Thanks!

matklad:
Ah, I think this doesn't actually work as we want it too :sob: 

I think we've lost dropchk here, no?

Before, we didn't have a custom drop, so a `Lazy<&'a T>` allowed dangling `'a` at the point where the Lazy is dropped. 

In this variation, we now have a custom drop, so that won't be allowed. There's a dropchk test for OnceCell, but not for lazy. 

Is there some way to have `#[may_dangle]` on stable?

matklad:
https://github.com/rust-lang/rust/issues/34761#issuecomment-1256691491

danielhenrymantilla:
Oh, good thing this wasn't released!

  - Note: let's try to add `arrrr` tests for `Lazy` too then 😅 

> Is there some way to have `#[may_dangle]` on stable?

Short answer is no; slightly longer answer is the post you've linked to, but which generalizes with big difficulty.

Even longer answer, which I was planning to write a crate and blog post about it, is that with HKTs/GATs it is possible to define an advanced API that would let people opt-into their specific usage `#[may_dangle]` at the cost of having to use that type.

So for `once_cell` this is a big no, since it would be a major usability regression, in comparison of which lack-of-covariance over `F` pales.

  - But for those curious, the API would be something like `LazyMayDangle<'a, Gat!(&'_ T)>`, with `Lazy<T>` then being some kind of alias for `LazyMayDangle<'static, Gat!(T)>`.

Conclusion (to share back to #167): blocked due to lack of `#[may_dangle]`: only `std`lib itself may realistically feature this on stable Rust, but `std`lib is precisely the area where things ought to remain somewhat KISS 😅

PR #236, opened 2023-06-03
Revert "Make `unsync::Lazy<T, F>` covariant in `F`"

danielhenrymantilla:
Indeed, #233 introduced a regression w.r.t `#[may_dangle]` 😔, apologies


matklad:
Absolutely no need to apologize here, it's very useful to know what we actually could do here (pick one of covariance/dropck). 

IF you don't want to see the work going to waste, consider submitting a similar PR to rust-lang/rust :0) For std, I think we must get both bits right, and there we #[may dangle].

PR #237, opened 2023-06-04
Remove `PhantomData<*mut Waiter>` so we don't need to manually `impl Send`

Enselic:
The effect of having `PhantomData<*mut Waiter>` is that the `OnceCell<T>` stops being `impl<T: Send> Send for OnceCell<T>`. But we want it to be `impl<T: Send> Send for OnceCell<T>`, because we manually mark it as such.

By removing the `PhantomData` we can also remove the manual `impl Send`. The net effect is zero changes to the public API surface (including auto traits) of `once_cell`:

```
$ cargo install cargo-public-api --locked
$ cargo public-api diff origin/master..remove-unneeded-phantomdata
Removed items from the public API
=================================
(none)

Changed items in the public API
===============================
(none)

Added items to the public API
=============================
(none)
```

A variant of this PR is to remove the `PhantomData` but keep the manual `impl Send`. Maybe you prefer the explicitness.

# Background of this PR

I am studying how `once_cell` is implemented, and could not at first understand why we needed the `PhantomData<*mut Waiter>`. Digging deeper, I concluded that we do not in fact need it.

### DIFF
--- a/src/imp_std.rs
+++ b/src/imp_std.rs
@@ -5,7 +5,6 @@
 
 use std::{
     cell::{Cell, UnsafeCell},
-    marker::PhantomData,
     panic::{RefUnwindSafe, UnwindSafe},
     sync::atomic::{AtomicBool, AtomicPtr, Ordering},
     thread::{self, Thread},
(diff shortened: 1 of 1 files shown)


matklad:
Yeah, phantom data is no longer needed, as we switched to AtomicPtr a while ago. Let’s keep the explicit Send bound here — send/sync bounds are versus subtle here, so better to spell them out explicitly

matklad:
Let me merge as is while I am at a laptop..

PR #246, opened 2023-08-10
Support slices in `OnceRef` and `OnceBox`

nvzqz:
This is implemented through a new public-in-private `OncePointee` trait. If this was instead done with `impl OnceRef<[T]>`, the compiler fails to infer which methods we want to use.

Because of the usage of `core::ptr::slice_from_raw_parts` in `const`, this change increases the MSRV to `1.64.0`.

To make correctness easier to reason about, I did not use `Ordering::Relaxed`. However, it's likely that this code could be improved by it.

### DIFF
--- a/src/race.rs
+++ b/src/race.rs
@@ -29,7 +29,7 @@ use atomic::{AtomicPtr, AtomicUsize, Ordering};
 use core::cell::UnsafeCell;
 use core::marker::PhantomData;
 use core::num::NonZeroUsize;
-use core::{mem, ptr};
+use core::ptr;
 
 /// A thread-safe cell which can be written to only once.
 #[derive(Default, Debug)]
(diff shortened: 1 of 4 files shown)


notgull:
I would discourage bumping the MSRV past 1.63, as that is the current MSRV for Debian Stable

matklad:
Compatibility with Debian stable is an explicit non-goal for once_cell:

https://github.com/matklad/once_cell/issues/201#issuecomment-1254883343

nvzqz:
I can rewrite this to use an associated type via `OncePointee` instead of `UnsafeCell<*mut T>`. That should remove the need to bump the MSRV.

src/race.rs
```diff
@@ -175,51 +176,275 @@ impl OnceBool {
...
+                Ok(_) => {
+                    // We initialized our length first, so store our pointer.
+                    once_ptr.store(ptr, Ordering::Release);
+                }
+                Err(real_len) => {
+                    len = real_len;
+
+                    // Spin briefly until the other thread initializes the
+                    // pointer.
+                    ptr = loop {
```
matklad, line 333: It’s a fundamental invariant of OnceCell that it doesn’t do unbounded spinning. If mutex-like blocking behavior is needed, then std dependency is required, because correct blocking fundamentally requires talking to the OS and notifying it which other thread you are blocked on, so that the OS can correctly guard against priority inversion.

matklad: My understanding is that no good solutions are available here: either a mutex or a cas is required. Though, I think Rust has double-word cas available on at least some architectures:

https://doc.rust-lang.org/stable/core/arch/x86_64/fn.cmpxchg16b.html

So I think the path forward here is to add support for DSTs via double-word-cas, by adding a bunch of cfgs everywhere. Eg, I'd imagine some 32 processors might have AtomicU64? Though, doing that probably requires a research of what's the status quo for wider-than-word atomic operations in today's Rust. That intrinsic exists, so its non-zero, but I don't know best practicies for actually using that.

matklad:
closing as per https://github.com/matklad/once_cell/pull/246#discussion_r1299254344

PR #247, opened 2023-08-21
Add MSRV to readme

Stargateur:
I think that good to put MSRV in the README.

Fix somehow https://github.com/matklad/once_cell/issues/201


Stargateur:
The only problem of putting it in the code is that user need to look rust documentation, I generally include the readme in the documentation crate level. Maybe you will be interested to include the readme in crate documentation ?

> This extra information is very useful for users who want to learn the context about a specific policy, but I don't think it's useful enough for enough of people to spend valuable README.md space on that.

Feel free to close to your project haha ^^ was just an idea

Stargateur:
I generally use that `#![doc = include_str!("../readme.md")]`

PR #248, opened 2023-09-20
add support for platforms without atomic ptr

LinuxHeki:
Add support for platforms without support for atomic operations on pointers (such as `thumbv6m-none-eabi`).

### DIFF
--- a/src/race.rs
+++ b/src/race.rs
@@ -54,6 +54,7 @@ impl OnceNonZeroUsize {
     ///
     /// Returns `Ok(())` if the cell was empty and `Err(())` if it was
     /// full.
+    #[cfg(target_has_atomic = "ptr")]
     #[inline]
     pub fn set(&self, value: NonZeroUsize) -> Result<(), ()> {
         let exchange =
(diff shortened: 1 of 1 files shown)


src/race.rs
```diff
@@ -64,6 +65,21 @@ impl OnceNonZeroUsize {
...
 
+    /// Sets the contents of this cell to `value`.
+    ///
+    /// Returns `Ok(())` if the cell was empty and `Err(())` if it was
+    /// full.
+    #[cfg(not(target_has_atomic = "ptr"))]
+    #[inline]
+    pub fn set(&self, value: NonZeroUsize) -> Result<(), ()> {
+        if self.inner.load(Ordering::Acquire) == 0 {
+            self.inner.store(value.get(), Ordering::Release);
```
matklad, line 76: Hm, this feels incorrect to me. 

Consider threads A, B, C and the following sequence of events:


```
   A   |   B     |   C
       |         |
load 0 |         |
       | load 0  |
       | store 1 |
       |         | load 1
store 2|         |
       |         | load 2
```

Here, A and B race when initializing the `inner`, so C observes two different values of once cell, which violates the invariant that makes the whole construct a safe abstraction.

LinuxHeki:
True... Does that mean this crate can't be ported platforms without atomic ptr?

taiki-e:
> platforms without atomic ptr

"critical-section" feature is a feature to use `sync`/`race` in such targets.
https://github.com/matklad/once_cell/blob/c48d3c2c01de926228aea2ac1d03672b4ce160c1/Cargo.toml#L48-L51

briansmith:
Should this be closed now?

PR #249, opened 2023-10-29
Remove one unsafe

joshlf:


matklad:
I’d say it’s not worth it. There are *much* scarier unsafe blocks in this crate already, and the safe code is way too verbose. Would be good to revisit once we can write `const {}` though!

PR #254, opened 2024-02-10
Add dependabot

oriontvv:
This PR adds dependabot config


matklad:
I don't think this is necessary for this project.

PR #260, opened 2024-09-02
propagate critical-section feature selection into portable-atomic; other minor updates for v1.20.0

### LINKED ISSUE #262: CI: detect `cargo fmt` issues

brodycj:
(updated)

- propagate `critical-section` feature selection into `portable-atomic`
- _update TEST_BETA - push RUSTFLAGS = `-A unreachable_patterns`_ - WORKAROUND ref: #261
- add `cargo test --workspace` to beginning of TEST task - resolves #262
- other minor updates
- bump minor version in Cargo.toml & update change log

CI build now fixed with TEMPORARY WORKAROUND in TEST_BETA task.

---

__ORIGINAL RATIONALE RE: propagate `critical-section` feature selection into `portable-atomic`__

Normally just using `critical-section` should be good enough for targets like super-outdated `thumbv6m-none-eabi` target. But in https://github.com/rustls/rustls/pull/2088 I would like to preserve existing use of `alloc` option, in which case I had to add explicit `portable-atomic` dependency with its `critical-section` feature enabled.

IMO it would be nice if I didn't have to add explicit sub-depencencies with this option. This is my proposal to enable `critical-section` for `portable-atomic` if `portable-atomic` is wanted.

Please let me know if you have any other idea or perspective concerning this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 - 
 
+## 1.20.0
+
+- propagate `critical-section` feature selection into `portable-atomic`; other minor updates [#260](https://github.com/matklad/once_cell/pull/260)
+
 ## 1.19.0
 
 - Use `portable-atomic` instead of `atomic-polyfill`, [#251](https://github.com/matklad/once_cell/pull/251).
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -21,6 +21,8 @@ fn main() -> xshell::Result<()> {
     {
         let _s = section("TEST");
 
+        cmd!(sh, "cargo test --workspace").run()?;
+
         for &release in &[None, Some("--release")] {
             cmd!(sh, "cargo test --features unstable {release...}").run()?;
             cmd!(
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -43,6 +43,11 @@ fn main() -> xshell::Result<()> {
     {
         let _s = section("TEST_BETA");
         let _e = push_toolchain(&sh, "beta")?;
+        // TEMPORARY WORKAROUND for Rust compiler issue ref:
+        // - https://github.com/rust-lang/rust/issues/129352
+        // - https://github.com/matklad/once_cell/issues/261
+        let _e = sh.push_env("RUSTFLAGS", "-A unreachable_patterns");
+
         cmd!(sh, "cargo test --features unstable").run()?;
     }
 
--- a/Cargo.lock.msrv
+++ b/Cargo.lock.msrv
@@ -43,7 +43,7 @@ checksum = "78ca9ab1a0babb1e7d5695e3530886289c18cf2f87ec19a575a0abdce112e3a3"
 
 [[package]]
 name = "once_cell"
-version = "1.19.0"
+version = "1.20.0"
 dependencies = [
  "critical-section",
  "parking_lot_core",
--- a/Cargo.lock.msrv
+++ b/Cargo.lock.msrv
@@ -69,6 +69,9 @@ name = "portable-atomic"
 version = "1.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7170ef9988bc169ba16dd36a7fa041e5c4cbeb6a35b76d4c03daded371eae7c0"
+dependencies = [
+ "critical-section",
+]
 
 [[package]]
 name = "redox_syscall"
--- a/Cargo.lock.msrv
+++ b/Cargo.lock.msrv
@@ -4,18 +4,18 @@ version = 3
 
 [[package]]
 name = "aho-corasick"
-version = "1.0.1"
+version = "1.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "67fc08ce920c31afb70f013dcce1bfc3a3195de6a228474e45e1f145b36f8d04"
+checksum = "8e60d3430d3a69478ad0993f19238d2df97c507009a52b3c10addcd7f6bcb916"
 dependencies = [
  "memchr",
 ]
 
 [[package]]
 name = "bitflags"
-version = "1.3.2"
+version = "2.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
+checksum = "b048fb63fd8b5923fc5aa7b340d8e156aec7ec02f0c78fa8a6ddc2613f6f71de"
 
 [[package]]
 name = "cfg-if"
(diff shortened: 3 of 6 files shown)


matklad:
Yes, this makes sense! Could you:

* bump the minor version in Cargo.toml
* add an entry to the changelog
* fix the CI (I think you need to update the Cargo.lock.msrv which we used to test old versions of Rust)

brodycj:
I just force-pushed an update to update the `Unreleased` section of `CHANGELOG.md`, will try to fix the CI, update the minor version, and add the minor version heading to `CHANGELOG.md` next. Thanks!

CHANGELOG.md
```diff
@@ -2,7 +2,10 @@
 
 ## Unreleased
 
-- 
+- propagate critical-section feature selection into portable-atomic
+- cleaner internal result unwapping (should help keep newer MIRI versions happy)
+- use `Option::unwrap_or_else` fn to avoid some unsafe blocks
+  (No is difference expected in terms of instructions executed, assuming unwrap is successful.)
```
brodycj, line 8:
```suggestion
  (No difference is expected in terms of instructions executed, assuming unwrap is successful.)
```

brodycj:
I have pushed some more updates that should resolve the CI build issues, bump the minor version, and update the change log.

I started using `Option::unwrap_or_else` as a cleaner way for result unwrapping to resolve the issues with recent MIRI versions & get rid of some `unsafe` blocks. I did use VS Code to inspect the Rust library code for `Option::unwrap_or_else` & `Option::unwrap_unchecked`; it looks to me like both cases are using `match` statements, with no difference in the expected `Some` arm.

I have also included a couple more minor updates:

- fix default-features flag for parking_lot_core in dependencies - as needed to resolve a cargo warning concerning deprecated use of `default_features` with underscore (`_`)
- used `cargo fmt` to fix a couple minor formatting issues
- add tidy cargo fmt check to xtask (main task)

@matklad it looks to me like there is quite a bit of duplicated code between the various OnceCell implementations; I am thinking it should be possible to reduce some of this duplicated code with some more extensive help from generics. I would like to try working on this kind of improvement if I can find some time, someday.

Many thanks for all of your efforts to offer & support this highly-useful package!

brodycj:
ping @matklad

brodycj:
@matklad any chance you can take a look?

src/lib.rs
```diff
@@ -587,11 +589,9 @@ pub mod unsync {
         where
             F: FnOnce() -> T,
         {
-            enum Void {}
-            match self.get_or_try_init(|| Ok::<T, Void>(f())) {
-                Ok(val) => val,
-                Err(void) => match void {},
-            }
```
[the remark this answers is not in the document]
brodycj: Not sure if I understand but I will remove updates like this.

> here is a bug in the complier: [rust-lang/rust#129031](https://github.com/rust-lang/rust/issues/129031)

Adding the following code to `lib.rs` DOES seem to resolve the MIRI errors:

```
#![allow(unreachable_patterns)]
```

Feels like this should be considered a *temporary* workaround that should be able to go away at some point in the future.

xtask/src/main.rs
```diff
@@ -77,6 +77,14 @@ fn main() -> xshell::Result<()> {
         cmd!(sh, "cargo miri test --features unstable").run()?;
     }
 
+    {
+        let _s = section("TIDY CARGO FMT CHECK");
+
+        sh.remove_path("./target")?;
+
+        cmd!(sh, "cargo fmt -- --check").run()?;
+    }
```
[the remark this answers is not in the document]
brodycj: I will put that command at the start of the TEST section.

src/imp_std.rs
```diff
@@ -68,7 +68,9 @@ impl<T> OnceCell<T> {
         initialize_or_wait(
             &self.queue,
             Some(&mut || {
-                let f = unsafe { f.take().unwrap_unchecked() };
```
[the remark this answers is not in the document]
matklad: The _purpose_ of `unwrap_unchecked` is to remove, at compile time, the possibility of the runtime panic, to make sure that the generated code doesn't contain panic invocation. In other words, this uses `unwrap_unchecked` specifically to avoid `.unwrap`, `unreacheable!()`, `panic!()` and frieds, to guarantee smaller codegen.

src/lib.rs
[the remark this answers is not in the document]
matklad: Let's avoid adding work-arounds for bugs in the _nightly_ versions of the compiler.

src/lib.rs
```diff
@@ -364,6 +364,8 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
+#![allow(unreachable_patterns)]
```
brodycj, line 367:
```suggestion
// NOTE: This is a temporary workaround for compiler issue:
// - https://github.com/rust-lang/rust/issues/129031
#![allow(unreachable_patterns)]
```

src/lib.rs
```diff
@@ -364,6 +364,10 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
+// NOTE: This is a temporary workaround for Rust compiler issue:
+// - https://github.com/rust-lang/rust/issues/129031
```
brodycj, line 368:
```suggestion
// NOTE: This is a temporary workaround for Rust compiler issue ref:
// - https://github.com/rust-lang/rust/issues/129031
// - https://github.com/rust-lang/rust/issues/129352
```

src/lib.rs
```diff
@@ -364,6 +364,10 @@
...
+#![allow(unreachable_patterns)]
```
brodycj, line 369: I am thinking it may be better to keep this per match statement where this is needed, along with these comments (should probably reference both rust-lang issues), in a similar fashion to this update on `async-lock`: https://github.com/smol-rs/async-lock/pull/92

brodycj:
@matklad I think this should be ready now. A couple minor things I can try to fix next week if you like:

- I did add `cargo test --workspace` to the beginning of TEST task as discussed, however this still does not seem to fail in case of formatting issues. I already raised issue #262 to track this, would be happy to investigate in the near future. But I would kinda rather not block this update for #262.
- I think I should have added `#![allow(unreachable_patterns)]` per `match` statement where this is needed, in a similar fashion to a similar update on `async-lock`: https://github.com/smol-rs/async-lock/pull/92 ... I can try to fix this in the near future if you like.

Thanks!

src/lib.rs
matklad: I don't think that's the right approach -- this crate does not officially support nightly Rust, and I don't want to pick up maintainance burden to be compatible with nightlies. 

So, there should not be source-level work-arounds for nightly bugs.

brodycj: How should we move forward then?

brodycj: I think using `RUSTFLAGS=\"-A unreachable_patterns\"` should be able to help, trying in my own fork for now

brodycj: Using `sh.push_env("RUSTFLAGS", "-A unreachable_patterns");` in the `TEST_BETA` does seem to fix the CI building without affecting any of the source code, and I did make a similar comment in issue #261.

brodycj:
@matklad I think this should be ready (yet again), with a TEMPORARY WORKAROUND in I made in the `TEST_BETA` task. Please let me know if you need anything else for this. Thanks!

xtask/src/main.rs
```diff
@@ -41,6 +43,11 @@ fn main() -> xshell::Result<()> {
     {
         let _s = section("TEST_BETA");
         let _e = push_toolchain(&sh, "beta")?;
+        // TEMPORARY WORKAROUND for Rust compiler issue ref:
+        // - https://github.com/rust-lang/rust/issues/129352
+        // - https://github.com/matklad/once_cell/issues/261
+        let _e = sh.push_env("RUSTFLAGS", "-A unreachable_patterns");
```
matklad, line 49: Ahh, I didn't realize this percolated all the way up to the beta! But, anyway, you've found the _perfect_ solution here, which places the fine-grained work-around where it belongs, thanks a bunch for this!

matklad (commented):
Looks awesome to me now, thanks for persevering here!

brodycj:
🎉

kaspar030:
> * propagate `critical-section` feature selection into `portable-atomic`

This breaks our use case. `portable-atomic` treats `critical-section` and `unsafe-assume-single-core` as mutually exclusive. E.g., we depend on esp-rs, which for some platforms (e.g., esp32c2, c3) explicitly enables the latter.

Could we make this optional?

brodycj:
I would like to state my assumption that despite issue #264, all other changes from this PR should be considered valid and should remain in `master` (and be part of the next release):

- cargo fmt updates that would be needed to pass `cargo fmt --check`
- fix default-features flag for parking_lot_core in dependencies (in response to a warning message I noticed)
- CI-related updates:
  - add `cargo test --workspace` to beginning of TEST task (as discussed above, this is needed to ensure we check formatting in CI workflow)
  - temporary workaround for Rust compiler bug in RUSTFLAGS, as was needed for `TEST_BETA` task to succeed

matklad please comment in case this is mistaken ... apologies for commenting on an already merged & closed PR!

PR #265, opened 2024-09-22
Add portable-atomic feature and disable portable-atomic/critical-section by default

### LINKED ISSUE #264: regression with portable-atomics in `1.20.0`
#260 propagates the `critical-section` feature to portable-atomics, which conflicts with `unsafe-assume-single-core`.

> * propagate `critical-section` feature selection into `portable-atomic`

This breaks our use case. `portable-atomic` treats `critical-section` and `unsafe-assume-single-core` as mutually exclusive. We depend on esp-rs, which for some platforms (e.g., esp32c2, c3) explicitly enables the latter.

It is my understanding that `libraries` should not set either `critical-section` or `unsafe-assume-single-core`, but leave that to the end user.

#260 states this:

```
ORIGINAL RATIONALE RE: propagate critical-section feature selection into portable-atomic

Normally just using critical-section should be good enough for targets like super-outdated thumbv6m-none-eabi target. But in https://github.com/rustls/rustls/pull/2088 I would like to preserve existing use of alloc option, in which case I had to add explicit portable-atomic dependency with its critical-section feature enabled.

IMO it would be nice if I didn't have to add explicit sub-depencencies with this option. This is my proposal to enable critical-section for portable-atomic if portable-atomic is wanted.

Please let me know if you have any other idea or perspective concerning this.
```

What does `critical-section` on `portable-atomics` actually enable that only `critical-section` feature enable? Maybe this can be solved in `portable-atomic` with a backend-agnostic feature like `require-cas`?

taiki-e:
Closes #264

This implements the first of the ways I mentioned in https://github.com/matklad/once_cell/issues/264#issuecomment-2354148809.

> * New Cargo feature to use `portable-atomic`'s atomic types instead of `core`'s

(The second way mentioned there was probably more user-friendly, but I did not choose it because it would have caused `portable-atomic` to appear in Cargo.lock as a platform-specific dependency. Like cfg-specific dependencies in https://github.com/crossbeam-rs/crossbeam/pull/487#issuecomment-783365041.)

No new test case was added, because the following case is sufficient to test the situation where the `race` module is enabled (due to default features) and `portable-atomic` is used instead of `core::sync::atomic` (due to `critical-section` feature enables `portable-atomic` feature).

https://github.com/matklad/once_cell/blob/72f7c2e5faed09e84aa1ffe26c25092c87954b98/xtask/src/main.rs#L40

This also closes #248 since this is a sound way to do that.

cc @kaspar030 @brodycj

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ members = ["xtask"]
 
 [dependencies]
 parking_lot_core = { version = "0.9.10", optional = true, default-features = false }
-portable-atomic = { version = "1.7", optional = true }
+portable-atomic = { version = "1.7", optional = true, default-features = false }
 critical-section = { version = "1.1.3", optional = true }
 
 [dev-dependencies]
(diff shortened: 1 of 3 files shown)


matklad:
This all looks alright to me, thanks a lot for putting this together!

Cargo.toml
```diff
@@ -38,17 +38,23 @@ std = ["alloc"]
 alloc = ["race"]
 
 # Enables `once_cell::race` module.
-race = []
+race = ["portable-atomic?/require-cas"]
```
taiki-e, line 41: Hmm, due to Cargo bug (https://github.com/rust-lang/cargo/issues/10801), this makes portable-atomic to appear in Cargo.lock (even if the portable-atomic feature is not enabled. even if portable-atomic will never be compiled).

With the diagnostic improvements in portable-atomic 1.8 (https://github.com/taiki-e/portable-atomic/pull/181), the need for require-cas feature has decreased, so I think it is probably fine to remove `"portable-atomic?/require-cas"`...

matklad: I am also fine with keeping it! Cargo bug is a bug, the code here seems correct, and could perhaps be helpful for older versions of portable-atomic? 

I guess you are the best person to make a judgement call here though :)

brodycj:
Thanks @taiki-e & @matklad & thanks for the quick release!

PR #269, opened 2024-12-24
Add get_or_default for OnceCell

loongs-zhang:


matklad:
I’d rather not add it, as this adds nothing of essence on top of `self.get_or_init(T::default)` and instead creates ambiguity. A method with this name could also be implemented as `get().unwrap_or_default()`

PR #273, opened 2025-02-06
Make initialization in `OnceNonZeroUsize::get_or_try_init` `#[cold]`.

briansmith:
In typical use, the Some branch is going to be taken extremely frequently but the None branch is only going to be taken once (or a very small number of times) at startup. If `get_or_try_init` is in a performance-sensitive segment of code, then it is important that `get_or_try_init` be inlined and that the compiler understands that the Some branch is (much) more likely than the None branch. When this happens, the call site basically becomes a load followed by a conditional jump that is basically never taken; which is ideal.

When `get_or_try_init` is used in many places in the user's code, it is important to avoid inlining any of the None branch into the call sites.

Unfortunately, the Rust compiler is sometimes not good at recognizing that code that calls a #[cold] function unconditionally must be cold itself. So, sometimes it isn't enough to mark our f as `#[cold] #[inline(never)]`.

Move the entire body of the None branch into a function that is marked because some post-inlining optimization passes in the compiler seem to not understand `#[cold]`, and because we don't want any part of that branch to be in the calling code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Outline initialization in `race`: [#273](https://github.com/matklad/once_cell/pull/273).
+
 ## 1.20.2
 
 - Remove `portable_atomic` from Cargo.lock if it is not, in fact, used: [#267](https://github.com/matklad/once_cell/pull/267)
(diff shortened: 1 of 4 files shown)


matklad (approved):

matklad:
Thanks!

PR #274, opened 2025-02-06
Add `OnceNonZeroUsize::get_unchecked`.

briansmith:
Each call to `OnceNonZeroUsize::get()` results in an atomic load as the compiler will not cache and reuse the result of an atomic load.

Provide an analogous `get_unchecked()` that uses a non-atomic load. The compiler will be able to cache and reuse the result to avoid redundant loads.

### DIFF
--- a/src/race.rs
+++ b/src/race.rs
@@ -50,6 +50,31 @@ impl OnceNonZeroUsize {
         NonZeroUsize::new(val)
     }
 
+    /// Get the reference to the underlying value, without checking if the cell
+    /// is initialized.
+    ///
+    /// # Safety
+    ///
+    /// Caller must ensure that the cell is in initialized state, and that
+    /// the contents are acquired by (synchronized to) this thread.
+    pub unsafe fn get_unchecked(&self) -> NonZeroUsize {
+        let p = self.inner.as_ptr();
+
+        // SAFETY: The caller is responsible for ensuring that the value
+        // was initialized and that the contents have been acquired by
+        // this thread. Assuming that, we can assume there will be no
+        // conflicting writes to the value since the value will never
+        // change once initialized. This relies on the statement in
+        // https://doc.rust-lang.org/1.83.0/core/sync/atomic/ that "(A
+        // `compare_exchange` or `compare_exchange_weak` that does not
+        // succeed is not considered a write."
+        let val = unsafe { p.read() };
+
+        // SAFETY: The caller is responsible for ensuring the value is
+        // initialized and thus not zero.
+        unsafe { NonZeroUsize::new_unchecked(val) }
+    }
+
     /// Sets the contents of this cell to `value`.
     ///
     /// Returns `Ok(())` if the cell was empty and `Err(())` if it was


briansmith:
Wow. `AtomicUsize::as_ptr()` wasn't added until Rust 1.70; see
https://doc.rust-lang.org/std/sync/atomic/struct.AtomicUsize.html#method.as_ptr.

matklad:
It's fine to bump rust-version to 1.70! We support at least 8 past versions!

briansmith:
The comment needs to be updated to directly address this, and probably other aspects of the memory model: "The most important aspect of this model is that data races are undefined behavior. A data race is defined as conflicting non-synchronized accesses where at least one of the accesses is non-atomic. Here, accesses are conflicting if they affect overlapping regions of memory and at least one of them is a write. They are non-synchronized if neither of them happens-before the other, according to the happens-before order of the memory model." (from https://doc.rust-lang.org/std/sync/atomic/index.html).

briansmith:
> It's fine to bump rust-version to 1.70! We support at least 8 past versions!

My project doesn't yet, though! :)

src/race.rs
```diff
@@ -50,6 +50,30 @@ impl OnceNonZeroUsize {
...
+    pub fn get_unchecked(&self) -> NonZeroUsize {
+        let p = self.inner.as_ptr();
+
+        // SAFETY: The caller is responsible for ensuring that the value
+        // was initialized and that the contents have been acquired by
+        // this thread. This method is only made available on a subset of
+        // targets where we are confident there will be no tearing or other
+        // interference with another thread that may be setting the value
+        // simultaneously.
+        let val = unsafe { p.read() };
```
matklad, line 70: I think non-atomic read will be correct on any platform? The cell _must_ be initialized at this point, so, even if there are some concurrent `self.inner.compare_exchange`, they all will fail, and won't cause teared reads.

You can get the tear if you don't actually ensure the safety invariant here, and do `set/get_unchecked` literally in parallel, but that would be technical UB on x86 as well, from the perspective of Rust abstract machine.

src/race.rs
```diff
@@ -50,6 +50,30 @@ impl OnceNonZeroUsize {
...
 
+    /// Get the reference to the underlying value, without checking if the cell
+    /// is initialized.
+    ///
+    /// # Safety
+    ///
+    /// Caller must ensure that the cell is in initialized state, and that
+    /// the contents are acquired by (synchronized to) this thread.
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    pub fn get_unchecked(&self) -> NonZeroUsize {
```
matklad, line 61:
```suggestion
    pub unsafe fn get_unchecked(&self) -> NonZeroUsize {
```

The important part!

matklad:
>My project doesn't yet, though! :)

I am fine with this PR applying some work-around to get it work on older compilers, but, in general no promises that future versions of OnceCell will support 1.70.0! While I won't proactively bump to stable-8 just because, I will bump as soon as compat becomes inconvenient

src/race.rs
briansmith: > even if there are some concurrent self.inner.compare_exchange, they all will fail, and won't cause teared reads.

I will think about this one overnight. I have MSRV 1.63.0 in my projects so I might try to find an alternative.

src/race.rs
briansmith: Haha. Yes, I also need to add some tests and improve the documentation.

src/race.rs
matklad: I wonder if just doing Relaxed load could optimize well?

briansmith: I tested it in my codebase. The relaxed load generates the same exact code as using `get()`. The code as presented here avoids subsequent reloads after the first call to `get()`.

briansmith:
Comparing https://doc.rust-lang.org/1.74.0/core/sync/atomic/index.html and https://doc.rust-lang.org/1.75.0/core/sync/atomic/index.html shows that 1.75.0 may be the earliest version where we can fully reason about atomics and UB, as that was the first version that attempted to document the memory model for them.

briansmith:
Also, https://doc.rust-lang.org/1.83.0/core/sync/atomic/index.html is the first version that talks about conflicting accesses. It says

> The most important aspect of this model is that data races are undefined behavior. A data race is defined as conflicting non-synchronized accesses where at least one of the accesses is non-atomic. Here, accesses are conflicting if they affect overlapping regions of memory and at least one of them is a write. They are non-synchronized if neither of them happens-before the other, according to the happens-before order of the memory model.

That leads to the question, is a `compare_exchange` considered a "write" if it doesn't succeed in changing the value? It isn't clear.

briansmith:
https://github.com/rust-lang/rust/issues/136669

briansmith:
According to https://doc.rust-lang.org/1.60.0/core/sync/atomic/struct.AtomicUsize.html, `AtomicUsize` is guaranteed to have the same layout and bit validity as usize even as far back as 1.60 (maybe further, didn't check), which is OnceCell's MSRV. 

In later documentation, it is clarified that the alignment of `AtomicUsize` may be larger (equal to its size) on targets where usize has lower alignment. Thus I think we can assume `alignof::<AtomicUsize>() >= alignof::<usize>()` (and we can check this with a const assertion). Thus, I believe we have enough guaranteed to implement a polyfill for `as_ptr()`.

matklad:
Yes, I agree with this reasoning, it should be valid to cast a pointer to atomic usize to a pointer to a normal usize

briansmith:
Just FYI, I was able to implement a workaround to deal with the added reloading when switching from `spin::Once::get_unchecked` to `OnceNonZeroUsize::get()`. Basically my calls to `get()` were done in batches, and I introduced an intermediate variable that held the loaded value, and then did the batch of checks on the new intermediate value, instead of calling `get()` over and over again.

Nonetheless, I am open to continuing to work on this.

RalfJung:
Note that that is a docs-only change. So it does apply retroactively, in a sense.

However it'd be good to wait until it actually lands, just to be sure. (In theory it could get backed out again but that seems unlikely.)

src/race.rs
```diff
@@ -50,6 +50,30 @@ impl OnceNonZeroUsize {
...
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    pub unsafe fn get_unchecked(&self) -> NonZeroUsize {
+        let p = self.inner.as_ptr();
+
+        // SAFETY: The caller is responsible for ensuring that the value
+        // was initialized and that the contents have been acquired by
+        // this thread. This method is only made available on a subset of
+        // targets where we are confident there will be no tearing or other
+        // interference with another thread that may be setting the value
+        // simultaneously.
```
RalfJung, line 69: FWIW it is basically never correct in Rust to do target-specific reasoning like that, except when explicitly documented on our side.

Either there are data races, then it's UB everywhere, or there are no data races, then it's fine everywhere.

briansmith:
I have updated this PR based on the above comments (thanks RalfJung).

briansmith:
I filed a follow-up at https://github.com/rust-lang/rust/issues/138246.

Also, we still have the issue that `as_ptr()` is not implemented until Rust 1.70. Are we confident about casting the address of the AtomicUsize to a `*const usize` for the polyfill?

briansmith:
Note that the MSRV check is failing because of the use of `as_ptr`:
```
error[E0599]: no method named `as_ptr` found for struct `AtomicUsize` in the current scope
    --> src/race.rs:61:28
     |
  61 |         let p = self.inner.as_ptr();
     |                            ^^^^^^ method not found in `AtomicUsize`
  
  For more information about this error, try `rustc --explain E0599`.
```
Did you intend to merge this?

matklad:
Life without bors is hard! I did intend to merge this, but I certainly hoped that tests would not fail!

I think it's fine to use `1.70` as MSRV, and I'd be willing to accept the patch lowering it via replacing as_ptr with a cast, which I am sure should be fine!

PR #275, opened 2025-03-07
feat: impl `Clone for OnceBox` and `OnceBox::with_value`

klkvr:
Those seem useful, unsure if there's a reason those were not provided right away


matklad:
Makes sense to me! Could you also add a line to the changelog and bump version in Cargo.toml to release this immediately?

Issue #220, opened 2023-02-14
Potential to weaken ordering for success case of `compare_exhange`s in `race` module


Imberflur:
For example, the compare exchange for `OnceNonZeroUsize` uses `AcqRel` for the success case:
https://github.com/matklad/once_cell/blob/d706539c6f7cd47d0e8037d832c0c95214fb5f91/src/race.rs#L100

The `Acquire` portion of the `AcqRel` here is for the load of `0`, however there is no `Release` store of `0` so this will not synchronize with anything. Note that the construction of `OnceNonZeroUsize` still has a happens-before relationship to this since we have an `&self` reference. Thus, I think a `Release` ordering is sufficient here for the desired synchronization to be achieved ([including taking this documentation into account](https://docs.rs/once_cell/1.17.0/once_cell/race/index.html#atomic-orderings)):
```rs
 self.inner.compare_exchange(0, val, Ordering::Release, Ordering::Acquire);
````

I believe this applies to all uses of `compare_exchange` in this module.

----
(happy to make a PR if this looks reasonable)

matklad:
This _does_ look reasonable to me, yeah! Although, as usual, I am unnerved by the fact that we don't have any tools to check this (or maybe I am missing them?).

Could you send a PR? You can also adjust the Changelog and bump the patch version in Cargo.toml, so that we release this together with #219

taiki-e:
Note that CAS failure ordering that is stronger than success ordering requires Rust 1.64+: https://github.com/rust-lang/rust/pull/98383

Imberflur:
> Note that CAS failure ordering that is stronger than success ordering requires Rust 1.64+: https://github.com/rust-lang/rust/pull/98383

Gah! :slightly_frowning_face:

> Although, as usual, I am unnerved by the fact that we don't have any tools to check this (or maybe I am missing them?).

The only thing that comes to my mind is testing with [loom](https://docs.rs/loom). I don't know if it is exactly what you are looking for though.

briansmith:
Does anybody know of how the codegen would change such that this would result in better performance for users of the race module?

briansmith:
I believe it is good to raise MSRV to 1.64 so we can do this. But it would be good to know what the real benefit would be.

Imberflur:
FWIW I don't have any specific motivating case for this. I asked about this on the community discord and someone mentioned:
> so in theory llvm can optimize it in some additional ways, but it doesn't in simple cases

[...]

PR #277, opened 2025-03-13
Reduce MSRV to 1.65 by polyfilling `AtomicUsize::as_ptr()`.

briansmith:
Potentially the MSRV could be reduced further as `cargo +1.63.0 check` passes, but `cargo +1.63.0 test` fails:

```
$ cargo +1.63.0 test
error: package `regex-syntax v0.8.5` cannot be built because it requires
rustc 1.65 or newer, while the currently active rustc version is 1.63.0
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,14 @@
 # Changelog
 
+## 1.21.1
+- Reduce MSRV to 1.65: [#277](https://github.com/matklad/once_cell/pull/277).
+
 ## 1.21.0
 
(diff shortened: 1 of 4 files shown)


briansmith:
I have updated the PR to set MSRV to 1.65 since that is the minimum for `cargo test` to work, because I think any MSRV <= 1.66 is good enough for almost anybody, and because I think we'll benefit from at least MSRV 1.64 by fixing issue #220.

matklad:
Thanks!

PR #278, opened 2025-03-13
race: Relax success ordering from `AcqRel` to `Release`.

briansmith:
See the analogous change in
https://github.com/rust-lang/rust/pull/131746 and the discussion in
https://github.com/matklad/once_cell/issues/220.

What is the effect of this change? Not much, because before we ever execute the `compare_exchange`, we do a load with `Ordering::Acquire`; the `compare_exchange` is in the `#[cold]` path already. Thus, this just mostly clarifies our expectations.  See the non-doc comment added under the module's doc comment for the reasoning.

How does this change the code gen? Consider this analogous example:

```diff
 #[no_mangle]
 fn foo1(y: &mut i32) -> bool {
-    let r = X.compare_exchange(0, 1, Ordering::AcqRel, Ordering::Acquire).is_ok();
+    let r = X.compare_exchange(0, 1, Ordering::Release, Ordering::Acquire).is_ok();
     r
 }

```

On x86_64, there is no change. Here is the generated code before and after:

```
foo1:
        mov     rcx, qword ptr [rip + example::X::h9e1b81da80078af7@GOTPCREL]
        mov     edx, 1
        xor     eax, eax
        lock            cmpxchg dword ptr [rcx], edx
        sete    al
        ret

example::X::h9e1b81da80078af7:
        .zero   4
```

On AArch64, regardless of whether atomics are outlined or not, there is no change. Here
is the generated code with inlined atomics:

```
foo1:
        adrp    x8, :got:example::X::h40b04fb69d714de3
        ldr     x8, [x8, :got_lo12:example::X::h40b04fb69d714de3]
.LBB0_1:
        ldaxr   w9, [x8]
        cbnz    w9, .LBB0_4
        mov     w0, #1
        stlxr   w9, w0, [x8]
        cbnz    w9, .LBB0_1
        ret
.LBB0_4:
        mov     w0, wzr
        clrex
        ret

example::X::h40b04fb69d714de3:
        .zero   4
```

For 32-bit ARMv7, with inlined atomics, the resulting diff in the object code is:

```diff
@@ -10,14 +10,13 @@
         mov     r0, #1
         strex   r2, r0, [r1]
         cmp     r2, #0
-        beq     .LBB0_5
+        bxeq    lr
         ldrex   r0, [r1]
         cmp     r0, #0
         beq     .LBB0_2
 .LBB0_4:
-        mov     r0, #0
         clrex
-.LBB0_5:
+        mov     r0, #0
         dmb     ish
         bx      lr
 .LCPI0_0:
@@ -54,4 +53,3 @@

 example::X::h47e2038445e1c648:
         .zero   4
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 1.21.2
+- Relax success ordering from AcqRel to Release in `race`: [#278](https://github.com/matklad/once_cell/pull/278).
+
 ## 1.21.1
 - Reduce MSRV to 1.65: [#277](https://github.com/matklad/once_cell/pull/277).
 
--- a/Cargo.lock.msrv
+++ b/Cargo.lock.msrv
@@ -43,7 +43,7 @@ checksum = "78ca9ab1a0babb1e7d5695e3530886289c18cf2f87ec19a575a0abdce112e3a3"
 
 [[package]]
 name = "once_cell"
-version = "1.21.1"
+version = "1.21.2"
 dependencies = [
  "critical-section",
  "parking_lot_core",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "once_cell"
-version = "1.21.1"
+version = "1.21.2"
 authors = ["Aleksey Kladov <aleksey.kladov@gmail.com>"]
 license = "MIT OR Apache-2.0"
 edition = "2021"
--- a/src/race.rs
+++ b/src/race.rs
@@ -19,6 +19,14 @@
 //! `Acquire` and `Release` have very little performance overhead on most
 //! architectures versus `Relaxed`.
 
+// The "atomic orderings" section of the documentation above promises
+// "happens-before" semantics. This drives the choice of orderings in the uses
+// of `compare_exchange` below. On success, the value was zero/null, so there
+// was nothing to acquire (there is never any `Ordering::Release` store of 0).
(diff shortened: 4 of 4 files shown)


briansmith:
TODO:
- [ ] I noticed Aarch64 is the same before/after primarily because it has +outline-atomics by default whereas 32-bit ARM doesn't. So we might experiment with disabling that for Aarch64 to see more interesting results for that configuration.
- [ ] Instead of using the godbolt example from issue #220 in the commit message, we should probably document the difference from a real use of `get_or_init` as I did in https://github.com/briansmith/ring/pull/2331.

matklad:
LGTM, especially with the same change in Rust!

Could your remove the comments though, and instead add the (great) explanation of "there's nothing to synchronize with for `0`" to commit message? I think it's ok to hide this in git blame!

And, as usual, could you bump a version in Cargo.toml&add a changelog entry? ^^

briansmith:
> Could your remove the comments though, and instead add the (great) explanation of "there's nothing to synchronize with for 0" to commit message? I think it's ok to hide this in git blame!

I think it's too important to hide in the git log, especially when people go to change the code later. Instead, I added a non-doc comment under the doc comment, and removed the other comments I added.

> And, as usual, could you bump a version in Cargo.toml&add a changelog entry? ^^

done.

briansmith:
@matklad WDYT about this version? I have another proposed addition to once_cell that builds on this, so I'd like to make progress on deciding what to do about this.

matklad:
Sorry, just missed notification a week ago, thanks for the ping!

PR #282, opened 2025-03-27
Add `racerelaxed::OnceNonZeroUsizeRelaxed`.

briansmith:
Add a variant of `OnceNonZeroUsize` that uses relaxed ordering, for reduced overhead, especially on ARM/AArch64 and other weakly-ordered systems.

This is pretty much straight copy-pasta from `race` to `racerelaxed`:
```
git difftool HEAD:src/race.rs src/racerelaxed.rs
git difftool HEAD:tests/it/race.rs tests/it/racerelaxed.rs
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1410,3 +1410,6 @@ pub mod sync {
 
 #[cfg(feature = "race")]
 pub mod race;
+
+#[cfg(feature = "race")]
+pub mod racerelaxed;
--- a/src/racerelaxed.rs
+++ b/src/racerelaxed.rs
@@ -1,0 +1,148 @@
+//! Thread-safe, non-blocking, "first one wins" flavor of `OnceCell` *without*
+//! Acquire/Release semantics.
+//!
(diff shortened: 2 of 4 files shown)


briansmith:
> This is pretty much straight copy-pasta from race to racerelaxed:

... after PR #280 and PR #281.

matklad:
Hm, I am a bit apprehensive here. My overall philosophy is to not include everything possible, and instead leave more niche use-cases to be open-coded by the user. 

For a relaxed ordering, I think it might be clearer as well to _not_ use a library for it? At the call-site, with `cell.get_or_init(...)`, you aren't actually _seeing_ that the load is relaxed there! So, open-coding compare-exchange would lead to a more readable code. 

So I am inclined to say that this is out of scope for once_cell!

briansmith:
Yeah, I don't disagree with that. Another issue, even with my own usage, is that I probably need `OnceNonZeroU8Relaxed` and other variants too.

PR #286, opened 2025-03-28
DRY `race::OnceRef::{get_or_try_init, set}`.

briansmith:
Centralize all the pointer type casting and RMW into a new `compare_exchange` function as was done in other types in the module. Leave the dereferencing of the pointer in `init()` as `set()` doesn't ever use the old value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 1.21.3
+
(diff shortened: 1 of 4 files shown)


briansmith:
I think this is the last one 😄.

matklad:
If it is last, lest bump patch in Cargo.toml!

briansmith:
> If it is last, lest bump patch in Cargo.toml!

Done.

